PR #37, opened 2015-10-09
Refactor of flexible_type <--> python code.

hoytak:
This pull request reworks the flexible_type translation code.  New features:

1. Speedup: The new code is between 10% and 4x faster in constructing
an SArray/SFrame, depending on type, and 2-4x faster in accessing
SArray/SFrame elements.  A future work is to apply these changes
to the pylambda workers.

2. Consistent type inference rules.  Type inference is now handled
consistently when multiple types are present, with rules that
cover all the corner cases. Previously, many of these corner cases
were not handled in any consistent way.

- Integers, floats, strings, datetime, dict, arrays, and image all
have one-to-one correspondence with a flexible_type.

- Lists are translated to a vector if all elements are numeric, otherwise
they are translated to a list.

- An empty list is treated as a list.

- None is translated to flex_undefined.

Now, when type inference is done, and multiple types appear in an input
(e.g. empty lists and numeric lists in the input to an SArray), then the
following rules goven what type changes can be made in inferring a common
type:

- Rule 1: Integer can be upgraded to a float.
- Rule 2: A numeric list and a vector together make a vector
- Rule 3: None does not affect the type.  I.e. [t1,t2,None, t3] has the same
infered type as [t1, t2, t3]
- Rule 4: If a list or vector is present, an empty list does not
change the infered type.
- Rule 5: A vector can be upgraded to a list.

If a type cannot be chosen that is consistent with these rules, then a TypeError
exception is raised.

3. Type casting -- now the behavior when calling sf.SArray(data, dtype = type)
follows the same rules as SArray.astype.  Previously, this was part way
between type inference (very strict) and astype(...) loose, allowing more type
casting but basically following its own rules.  The behavior is now consistent
with astype.

4.  Cleanup:  The same names used in C++ (flex_list, flex_vec, etc.) are now used in cython.

-------------------------------------------------------------------------------------------------

*Examples of type inference when a common type is expected.   *

The defintions used are:

```
IntegerValue: int, long, bool, any numpy integer/boolean type.
FloatValue: float, any numpy floating point type.
StringValue: str, unicode, any numpy string/unicode type.
DictValue: dict.
DatetimeValue: date or datetime.
AnyValue: Any of the above values.
FloatSequence: List/tuple/iterable of floats, non-integer array, any numpy float array.
FloatSequenceWithNAN: Same as above, but contains a NAN.
FloatSequenceWithNone: Same as above, but contains a None.
IntegerSequence: List/tuple/iterable of floats, integer array, any numpy integer array.
IntegerSequenceWithNAN: Same as above, but contains a NAN.
IntegerSequenceWithNone: Same as above, but contains a None.
EmptyFloatArray: Empty python / numpy array with float type.
EmptyIntegerArray: Empty python / numpy array with integer  type.
EmptyArray: Either of the above.
EmptySequence: Empty list or tuple.
BooleanSequence: List/tuple of bools, boolean typed numpy array.
StringSequence: List/tuple of strings or unicode; numpy array of str/unicode.
AnySequence: Any of the above.
```

```
[IntegerValue]                        --> int
[IntegerValue, IntegerValue]          --> int
[IntegerValue, FloatValue]            --> float
[IntegerValue, nan]                   --> float
[]                                    --> int
[None]                                --> int
[IntegerValue, nan]                   --> float
[IntegerValue, None, nan]             --> float
[IntegerValue, None, FloatValue]      --> float
[IntegerValue, None, FloatValue, nan] --> float
[StringValue]                         --> str
[StringValue, StringValue]            --> str
[StringValue, IntegerValue]           --> NoneType
[StringValue, FloatValue]             --> NoneType
[DictValue]                           --> dict
[DictValue, DictValue]                --> dict
[AnySequence, AnyValue]               --> NoneType
[AnySequence, AnyValue, AnySequence]  --> NoneType
[AnySequence, AnyValue, AnyValue]     --> NoneType
[DatetimeValue, StringValue]          --> NoneType
[DatetimeValue, IntegerValue]         --> NoneType
[DatetimeValue, FloatValue]           --> NoneType
[EmptySequence]                       --> list
[IntegerSequence]                     --> array
[IntegerSequenceWithNone]             --> list
[IntegerSequenceWithNAN]              --> array
[FloatSequence]                       --> array
[FloatSequenceWithNAN]                --> array
[FloatSequenceWithNone]               --> list
[EmptyIntegerArray]                   --> array
[EmptyFloatArray]                     --> array
[BooleanSequence]                     --> array
[StringSequence]                      --> list
[IntegerSequence, FloatSequence]      --> array
[IntegerSequence, FloatSequence]      --> array
[EmptySequence, EmptyFloatArray]      --> array
[EmptySequence, EmptyIntegerArray]    --> array
[EmptySequence, IntegerSequence]      --> array
[EmptySequence, FloatSequence]        --> array
[EmptySequence, EmptyFloatArray]      --> array
[EmptySequence, EmptyIntegerArray]    --> array
[EmptySequence, IntegerSequence]      --> array
[EmptySequence, FloatSequence]        --> array
[StringSequence, EmptyFloatArray]     --> list
[StringSequence, EmptyIntegerArray]   --> list
[StringSequence, IntegerSequence]     --> list
[StringSequence, FloatSequence]       --> list
```
(Note, Adding a None in any of these does not change it.)


esamanas:
So is this right?
[1,2,3] would be translated to a vector of int
[1,2,None] would be translated to a vector of int
[1,2.0,3] would be translated to a vector of float
[1,2,"3"] would stay a list

hoytak:
There isn't a "vector of int" -- only a vector of floats. 

[1,2,3] would be translated to a vector
[1,2,None] would be a list
[1,2.0,3] would be translated to a vector
[1,2,"3"] would stay a list

ylow:
Well. when you say "[1,2,3] would be translated to a vector of int", do you mean
```
gl.SArray([1,2,3])
```
or
```
gl.SArray([[1,2,3]])
```
In the former it is an SArray of int type, in the latter it is an SArray of list type where the content is a single list containing integers.

hoytak:
Actually, `gl.SArray([[1,2,3]])` is an SArray of vector type.

esamanas:
I'm talking purely a value within an SArray.

So vectors are still only floats? I'm afraid this change may break timeseries group then.

hoytak:
Updated the comment to reflect type inference rules.

hoytak:
Evan: this PR should not break backwards compatibility except for cleaning up consistency of the edge cases and expanding the handling of type casting cases.

oss_src/unity/lib/gl_sarray.cpp
```diff
@@ -28,9 +28,11 @@ static graphlab::mutex reader_shared_ptr_lock;
  */
 flex_type_enum infer_type_of_list(const std::vector<flexible_type>& vec) {
   std::set<flex_type_enum> types;
+  flex_type_enum last_type = flex_type_enum::UNDEFINED;
   for (const flexible_type& val: vec) {
-    if (val.get_type() != flex_type_enum::UNDEFINED) {
+    if(val.get_type() != last_type && val.get_type() != flex_type_enum::UNDEFINED) {
       types.insert(val.get_type());
+      last_type = val.get_type();
```
ylow: last_type doesn't actually do very much here. This is an optimization and should be documented as such.

oss_src/unity/python/sframe/test/test_sframe.py
```diff
@@ -1372,64 +1373,69 @@ def test_concat(self):
...
+        # self.__assert_concat_result_equal(result.sort('a'), expected_result.sort('a'), ['List of b'])
 
-        result = sf.groupby('a', aggregate.CONCAT('d'))
-        expected_result = SFrame({
-            'a': [1,2,3,4, 5],
-            'List of d': [[1,1,2,2],[1,3,3],[4],[2], []]
-            })
-        self.__assert_concat_result_equal(result.sort('a'), expected_result.sort('a'), ['List of d'])
 
+        # result = sf.groupby('a', aggregate.CONCAT('d'))
```
ylow: These set of commented out tests. Are they incorrect? In which case they should just be deleted.

oss_src/unity/lib/gl_sarray.cpp
hoytak: Sounds good, will do.

oss_src/unity/python/sframe/test/test_sframe.py
hoytak: That was an oversight.  I'll reenable them.

PR #40, opened 2015-10-13
Convenience operators for converting between sframe/sarray flavors.

hoytak:
Added convenience functions to allow converting between any flavor of
(gl_sarray, unity_sarray, sarray) and between any flavor of
(gl_sframe, unity_sframe, sframe).  To use, simply include gl_sarray
or gl_sframe, and then call to_sarray, to_unity_sarray, or
to_gl_sarray, with any of the sarray flavors to get to the appropriate
type. SFrame stuff behaves the same way.

### DIFF
--- a/oss_src/unity/lib/CMakeLists.txt
+++ b/oss_src/unity/lib/CMakeLists.txt
@@ -27,6 +27,7 @@ make_library(unity_core
     unity_odbc_connection.cpp
     image_util.cpp
     ../extensions/additional_sframe_utilities.cpp
+    sframe_sarray_type_conversions.cpp
   REQUIRES
     flexible_type
     pylambda table_printer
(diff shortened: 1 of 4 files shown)


ylow:
A lot of these can and should be covered by implicit casts. Some of them already are. Especially the to_gl_sarray and to_gl_sframe conversion ones.

Also, I will rather not have these functions at all if possible. unity_sframe and unity_sarray should be considered as implementation detail at this level and users of gl_sarray / gl_sframe should require them. (And I will like to reduce unity_sframe / unity_sarray over time).

hoytak:
Fair enough.  One of the current pain points about the API that we should work on is that it's tricky to remember how to convert between different array / sframe types, as the functions for doing so are difficult to remember / discover.  I'd be happy to move them to implicit casts.

esamanas:
What's the status of this? Are we awaiting more commits?

ylow:
I don't feel like we should have this. We should hide the unity_* objects as much as possible and should not be part of a public maintained interface at all.

PR #42, opened 2015-10-15
Improve logging when launching processes

esamanas:
When launching a process (unity_server or pylambda_worker in this
commit) make sure that if the process errors out, the error message is
delivered to the user in some form (either to terminal or log). Also
some commands in the top namespace to print the location of the log to
look for errors.

Testing was in the form of manual adversarial tests, as it involved
breaking a current installation by moving files around, or changing
other system stuff.

### DIFF
--- a/oss_src/process/process_unix.cpp
+++ b/oss_src/process/process_unix.cpp
@@ -119,11 +119,14 @@ bool process::popen(const std::string &cmd,
     return false;
   } else if(pid == 0) {
     //***In child***
-    if(child_write_fd > -1) {
-      dup2(write_fd, child_write_fd);
-    }
-    close(write_fd);
     close(read_fd);
+    if((child_write_fd > -1) && (write_fd != child_write_fd)) {
+      errno = 0;
+      if(dup2(write_fd, child_write_fd) != child_write_fd) {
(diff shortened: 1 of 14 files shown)


hoytak:
Looks like this conflicts with https://github.com/dato-code/SFrame/pull/43; let's get this one in first, and I can update mine.

oss_src/process/process_unix.cpp
```diff
@@ -119,11 +119,14 @@ bool process::popen(const std::string &cmd,
...
     //***In child***
-    if(child_write_fd > -1) {
-      dup2(write_fd, child_write_fd);
-    }
-    close(write_fd);
     close(read_fd);
+    if((child_write_fd > -1) && (write_fd != child_write_fd)) {
+      errno = 0;
+      if(dup2(write_fd, child_write_fd) != child_write_fd) {
+        logstream(LOG_ERROR) << "dup2 failed to duplicate fd!" << strerror(errno) << std::endl;
```
ylow: the log message is probably not safe inside of a vfork.

ylow: I might just assert(dup2(write_fd ....) != child_write_fd).

(the c assert as opposed to our ASSERT)

PR #47, opened 2015-10-19
ODBC Fix

### LINKED ISSUE #19: gl.connect_odbc() fails
Repro steps:
1. Start Python session
2. 
```
import graphlab as gl
db = gl.connect_odbc(<my awesome connection string>)
```

Expected Behavior:
- db should now be connected to database

Actual Behavior:
```
AttributeError: 'module' object has no attribute '_odbc_connection'
```

ylow:
Fixes the issue that
```
import sframe as sf
sf.connect_odbc(...) # fails

sf.SArray()
sf.connect_odbc(...) # works
```

Normally any use of SFrame.extensions will activate all the extension loading.
However, in this case somehow SFrame.extensions failed to load if it was
imported too early. Might have something to do with the change to relative
imports.

Fixes #19

### DIFF
--- a/oss_src/unity/python/sframe/data_structures/DBConnection.py
+++ b/oss_src/unity/python/sframe/data_structures/DBConnection.py
@@ -5,8 +5,6 @@ All rights reserved.
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 '''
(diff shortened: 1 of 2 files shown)


ylow:
(Also modified an early import in _model.py)

PR #49, opened 2015-10-20
Ssl certificate issues

ylow:
Fixes several SSL Certificate issues.
 - Made all CURL calls use Python's Certifi store. This allows us to unify: i.e. if awscli works, we should work too. 
 - Switch the DMLCIO S3 block reader to use https
 - Added better error reporting from DMLCIO S3 block reader so we know when there is a certificate validation problem.
 - Enable SSL certificate check in Windows

### DIFF
--- a/oss_src/unity/python/sframe/sys_util.py
+++ b/oss_src/unity/python/sframe/sys_util.py
@@ -65,6 +65,12 @@ def make_unity_server_env():
 
     ## set local to be c standard so that unity_server will run ##
     env['LC_ALL']='C'
+    # add certificate file
+    try:
+        import certifi
+        env['GRAPHLAB_FILEIO_ALTERNATIVE_SSL_CERT_FILE'] = certifi.where()
+    except:
+        pass
     return env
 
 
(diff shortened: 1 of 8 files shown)


oss_src/unity/python/sframe/sys_util.py
```diff
@@ -65,6 +65,12 @@ def make_unity_server_env():
 
     ## set local to be c standard so that unity_server will run ##
     env['LC_ALL']='C'
+    # add certificate file
+    try:
+        import certifi
+        env['GRAPHLAB_FILEIO_ALTERNATIVE_SSL_CERT_FILE'] = certifi.where()
```
wangping25: Where is this environment used?

ylow: The function is make_unity_server_env. This is used to startup GLC.

wangping25: Understand. Just curious where the GLC code that actually uses this environment variable.

ylow: See the rest of the PR.

wangping25: Got it. I always tripped by the fact that you strip out the 'GRAPHLAB_' from environment variable name and use it inside unity_server as I always to a global search of original environment variable name!

Issue #34, opened 2015-10-08
Memory Size Detection


ylow:
Currently, we try to autotune the amount of memory we use by detecting the amount of system memory. However, this is problematic when we are run inside of docker since we detect the total amount of memory on the system (via sysinfo), rather the amount of memory allocated to us. (See http://fabiokung.com/2014/03/13/memory-inside-linux-containers/, https://github.com/docker/docker/issues/12394)

As of docker 1.8 which is rather new, we should be able to look into /sys/fs/cgroups or /proc/self/cgroup or something like that. But not everyone will be on docker 1.8, so we need an API workaround as well.

 - Use /sys/fs/cgroups to determine our memory limit and use that
 - ?Alternatives?
 - Add an sframe.set_memory_limit function that auto-configures the memory limit based on what the user tells us. This is the ultimate fallback.

hoytak:
Can this issue be closed?

shardool-freightwalla:
@ylow  Can you please elaborate how we can explicitly mention the sframe memory usuage. Couldn't find any definition about it.

I am getting the below error when i try to filter an sframe on heroku server. Locally it works perfect. The sframe size is < 1MB. I am unable to recreate it on heroku as manually running the same code works on heroku bash, so cant provide a reproducible code to demonstrate.

heroku stack :

Linux 3.13.0-91-generic #138-Ubuntu SMP Fri Jun 24 17:00:34 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.4 LTS
Release:	14.04
Codename:	trusty
RAM 512 MB

     *** Error in `python': double free or corruption (out): 0x00007fac2389c030 ***

ps: The app was working fine when used graphlab Sframe , but the open source gives the memory error. Attached logs.
[sframe_server_1469266810.txt](https://github.com/turi-code/SFrame/files/379494/sframe_server_1469266810.txt)

Hoping that setting the memory would help here.

shardool-freightwalla:
Runtime config if that helps.

[...]

PR #51, opened 2015-10-20
Memory Limit tuning

ylow:
- Tweaked the memory limit tuning procedure a bit to be much more conservative to take into consideration that we may not be the only process on the system. (i.e. tune to half the system memory rather than all of the system memory)
 - Added support for detection of memory limit inside of a cgroup.

addresses #34

### DIFF
--- a/oss_src/minipsutil/minipsutil.c
+++ b/oss_src/minipsutil/minipsutil.c
@@ -305,7 +305,29 @@ uint64_t total_mem() {
     if (sysinfo(&info) != 0) {
         return 0;
     }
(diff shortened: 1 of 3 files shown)


esamanas:
Why do you cut the row size estimate in half? It seems that would make groupby/join/sort LESS conservative.

ylow:
It only affects groupby. Yeah this is all guesses at the moment. I just *felt* like the groupby max cells is a little... small. Though I also halved the total_mem so it actually maintains the cell count.

PR #54, opened 2015-10-21
Cleans up cython directory on configure.

ylow:
This is necessary because CMake for whatever
reason leaves the build scripts around even after cython files get
deleted. So we just delete the cython build directory during
configure and force those build scripts to be regenerated.

### DIFF
--- a/configure
+++ b/configure
@@ -304,7 +304,11 @@ fi
 set -e
 set -o pipefail
 
-
+# Delete the cython make files. If cython files get removed,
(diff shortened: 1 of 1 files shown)


ylow:
@hoytak

esamanas:
Won't this error out if those directories don't exist?

ylow:
Nope. Thats the "f" in rm -rf". Even if it doesn't exist it will be fine.

PR #57, opened 2015-10-21
Partial Column Materialization

ylow:
Added an optimization to only materialize columns which need materializing.

i.e.
```
        sf["new_column"] = ... something complicated ...
        sf.__materialize__()
```
the materialize call should only materialize the new column and do not have to rewrite the entire sframe.

### DIFF
--- a/oss_src/sframe_query_engine/planning/optimizations/general_union_project_transforms.hpp
+++ b/oss_src/sframe_query_engine/planning/optimizations/general_union_project_transforms.hpp
@@ -224,10 +224,16 @@ class opt_project_add_direct_source_tags : public opt_transform {
       auto nn = n->inputs[idx_1];
       
       if(nn->type == planner_node_type::SFRAME_SOURCE_NODE) {
-        input_mapping[i] = nn->any_p<sframe>("sframe").select_column(idx_2); 
+        auto sa = nn->any_p<sframe>("sframe").select_column(idx_2);
+        if (nn->p("begin_index") == 0 && nn->p("end_index") == sa->size()) {
+          input_mapping[i] = sa;
+        }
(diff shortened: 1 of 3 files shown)


oss_src/sframe_query_engine/planning/optimizations/general_union_project_transforms.hpp
```diff
@@ -224,10 +224,16 @@ class opt_project_add_direct_source_tags : public opt_transform {
       auto nn = n->inputs[idx_1];
       
       if(nn->type == planner_node_type::SFRAME_SOURCE_NODE) {
-        input_mapping[i] = nn->any_p<sframe>("sframe").select_column(idx_2); 
+        auto sa = nn->any_p<sframe>("sframe").select_column(idx_2);
+        if (nn->p("begin_index") == 0 && nn->p("end_index") == sa->size()) {
```
hoytak: Why is this condition needed? I'm worried about it cause if it isn't true, input_mapping isn't set.

ylow: If the SOURCE node involves a sub-selected range (begin row, end row), than the input-mapping is insufficient information (since the input-mapping only contains column-id to the actual sarray, and not the range).

ylow: As far as I can tell input_mapping only goes into the field "direct_source_mapping" which is just column_id --> sarray.

hoytak: Ah, I see what's going on.  Good catch.

PR #60, opened 2015-10-23
Filling out the gl_sframe/gl_sarry API

hoytak:
Added:
- Implicit casting operators from sframe/sarray to gl_sframe/gl_sarray.
- materialize_to_sframe and materialize_to_sarray.
- added contains_column and column_name(index).

### DIFF
--- a/oss_src/unity/lib/unity_sframe.cpp
+++ b/oss_src/unity/lib/unity_sframe.cpp
@@ -323,6 +323,23 @@ size_t unity_sframe::column_index(const std::string &name) {
   return sf->column_index(name);
 }
 
+std::string unity_sframe::column_name(size_t index) {
+  Dlog_func_entry();
+
+  auto sf = this->get_underlying_sframe();
+  std::string ret = sf->column_name(index);
+  DASSERT_TRUE(ret == column_names().at(index));
+  return ret;
+}
+
+
+bool unity_sframe::contains_column(const std::string& name) {
+  Dlog_func_entry();
+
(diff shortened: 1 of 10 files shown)


oss_src/unity/lib/unity_sframe.cpp
```diff
@@ -323,6 +323,23 @@ size_t unity_sframe::column_index(const std::string &name) {
   return sf->column_index(name);
 }
 
+std::string unity_sframe::column_name(size_t index) {
+  Dlog_func_entry();
+
+  auto sf = this->get_underlying_sframe();
```
ylow: This can be done without get_underlying_sframe. This will force materialization to get a column name.

oss_src/unity/lib/unity_sframe.cpp
```diff
@@ -323,6 +323,23 @@ size_t unity_sframe::column_index(const std::string &name) {
...
+  std::string ret = sf->column_name(index);
+  DASSERT_TRUE(ret == column_names().at(index));
+  return ret;
+}
+
+
+bool unity_sframe::contains_column(const std::string& name) {
+  Dlog_func_entry();
+
+  const auto& sf = this->get_underlying_sframe();
```
ylow: ditto.

oss_src/unity/lib/unity_sframe.cpp
hoytak: Gotcha.  Will fix.

PR #61, opened 2015-10-23
More certificate issues

ylow:
Seems like letting CERT_DIR be the default value can cause problems on
some earlier versions on Linux. So we just force it to the empty string
which will allow us to *only* use the certificates in certifi

### DIFF
--- a/oss_src/unity/python/sframe/sys_util.py
+++ b/oss_src/unity/python/sframe/sys_util.py
@@ -72,10 +72,12 @@ def make_unity_server_env():
     ## set local to be c standard so that unity_server will run ##
     env['LC_ALL']='C'
     # add certificate file
-    if 'GRAPHLAB_FILEIO_ALTERNATIVE_SSL_CERT_FILE' not in env:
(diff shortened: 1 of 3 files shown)


oss_src/unity/python/sframe/sys_util.py
```diff
@@ -72,10 +72,12 @@ def make_unity_server_env():
     ## set local to be c standard so that unity_server will run ##
     env['LC_ALL']='C'
     # add certificate file
-    if 'GRAPHLAB_FILEIO_ALTERNATIVE_SSL_CERT_FILE' not in env:
+    if 'GRAPHLAB_FILEIO_ALTERNATIVE_SSL_CERT_FILE' not in env and \
```
hoytak: Should this be "or" instead of "and"?

ylow: ah... yes

ylow: Ah no... "and" is right.
If you set either, we don't override the environment.

PR #63, opened 2015-10-23
Use relative path in file_util and gl_pickle tests

wangping25:
Use relative path in the following test:

   test_file_util.py
   test_gl_pickle.py


oss_src/unity/python/sframe/test/test_gl_pickler.py
```diff
@@ -9,8 +9,8 @@
 from sframe.util import cloudpickle
 
 import sframe as gl
```
ylow: This is no good too right?
it can't be import sframe.
See the other sframe/tests for how to do this.

wangping25:
Fixed!

PR #64, opened 2015-10-23
There was a regression in the python extension code

ylow:
Conversion of SArray or SFrames to a C++ variant_type ends up
serializing the entire SFrame / SArray as a list.

```
>>> g=SFrame({'a':[1,2,3]})
>>> extensions._demo_identity({'b':g})
will return
{'b':[{'a':1},{'a':2},{'a':3}]} instead of {'b':SFrame...}
```

### DIFF
--- a/oss_src/unity/extensions/internal_demo.cpp
+++ b/oss_src/unity/extensions/internal_demo.cpp
@@ -15,6 +15,10 @@ int _demo_addone(int param) {
   return param + 1;
 }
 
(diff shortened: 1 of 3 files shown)


ylow:
The fix is kinda ugly at this point. The variant code needs the same love as the flexible_type code

hoytak:
Will merge for now, but we should rewrite that code.  It's pretty brittle.

PR #71, opened 2015-10-30
Numerous Fixes from internal testing

ylow:
This is a long list of fixes made during internal testing.

## Some segfaulting Fixes

    - Eigen apparently wants Eigen::initParallel()
    - make some environment variable checks not end in null pointer error
    if environment variable not set.

## Random

 - Added another multinomial sampling function

## More SSL fixes

 - The certificates are not picked up correctly
   in the webstor code (which we should deprecate)

## Funny Termination Oddity

- The following code takes a very very long time to terminate (20s-ish) if
     you just put it in a file and run python on it. (not ipython).
     ```
     import sframe as gl
     gl.SArray([1,2]).apply(lambda x:x)
     ```
     and it will print out "Unable to reach server" a few times before
     terminating.

     Can't quite figure out why, but it looks like unity_server is waiting
     for ping replies from the lambda workers, and the ping replies are not
     coming. So it has to wait for full 5 seconds for timeout.

     What is doubly fun is that if you do this
     ```
     import sframe as gl
     gl.SArray([1,2]).apply(lambda x:x)
     raw_input("pika") # and you wait for one second before pressing enter.
     ```
     it works great.

     Fix
      - made the ping wait in 1 second increments rather than 5 seconds at
      a time. This speeds up the above case to something more like ... 6s
      for full termination. Which is a little better. But I can't quite
      figure out why the ping is not responding.
      - Removed the "Unable to reach" message.

## Log Message Printing

- Now that stderr is piped all the way to display, we have some new
  printing oddities. In a number of places, we use exceptional control
  flow and that is no good. i.e.
    ```
            bool try_to_open_file(const std::string url) {
              bool success = true;
              try {
                general_ifstream fin(url);
                success = !fin.fail();
              } catch(...) {
                success = false;
              }
              return success;
            }
    ```

    This means that when try_to_open_file(...) returns false, which is
    actually ok, it emits a bunch of ugly errors to screen. Like:
    ```
    1446060349 : ERROR:    union_fstream.cpp(operator():98): Cannot open /tmp/tmpAdJ8Q5/dir_archive.ini for reading
    1446060349 : ERROR:    general_fstream.cpp(operator():29): Cannot open /tmp/tmpAdJ8Q5/dir_archive.ini for read. Cannot open /tmp/tmpAdJ8Q5/dir_archive.ini for reading
    1446060349 : ERROR:    unity_global.cpp(operator():121): Unable to load model from /tmp/tmpAdJ8Q5: Cannot open /tmp/tmpAdJ8Q5/dir_archive.ini for read. Cannot open /tmp/tmpAdJ8Q5/dir_archive.ini for reading
    ```
    I have an incomplete fix for this case (in that it will still emit ugly
    messages if the file actually exists, but is unreadable). But this is
    something we need to be more careful about.

    Since this does appear in a number of places, for now, after offline
    discussion with Evan, we are bumping LOG_WARNING and LOG_ERROR to
    stdout, where it is *not* forwarded to the user (it will still show
    up in the log file).

    More generally, we need some better exception handling. i.e.

    try {
        log_and_throw(...);  // this does not print to display, but logs it
                             // The message gets attached to the exception
    }

    Then the user doesn't actually see LOG_ERROR messages but all of those
    messages just get associated with the exception.

    Also fixed some color printing issues (don't print the color codes
    if we are not going to actually print the message)

## Windows Linking

    There were 2 issues.
    First issue was with linking. Windows requires the dll (or a .lib) file
    to be present for linking; even if you are linking a shared library.
    This means the "python_header_only" dependency cannot exactly be "header
    only". But it is still fine.

    Second issue is with termination.
    Windows appears to kill threads then call DLLUnload (which calls the
    global object destructors). This is problematic since all of our
    state is in DLLs. We can do more cleanup here or we can just die.
    (this is going to be a problem when we eliminate the unity_server
    subprocess)

## Removing libpython dependency

     - Split the pylambda library into two halves (pylambda and
         pylambda_worker_lib).
       - pylambda which is the "server" only side only and doesn't know how
       to speak python (this is the commander side which just sends lambdas to be
       executed) , and pylambda_worker_lib which knows how to speak python (which
       actually executes the lambdas)

     - Also, everything links in boost python which causes issues.
       - Split boost into 2 halves, boost and boost_python
       - Add a python_header_only library

     - This required the addition of an "override_cmake" directory which
     contains rules that overrides the files in cmake/. This can be thought of as a
     staging area for rules to go before adding them to dato-deps (our dependency
     building repository)


ylow:
Also @hoytak

PR #75, opened 2015-11-03
Fix test for windows

ylow:
int and long are not so consistent going through the variant conversion.
But that's all right. Python silently upgrades int and longs.


oss_src/unity/python/sframe/test/test_extensions.py
```diff
@@ -14,7 +14,10 @@
 class VariantCheckTest(unittest.TestCase):
 
     def identical(self, reference, b):
-        self.assertEquals(type(reference), type(b))
+    	if type(reference) in [int, long]:
```
esamanas: fyi you can also use six.integer_types. Not enough to gate this though.

PR #81, opened 2015-11-04
GLPickle expandvars and expanduser paths are now supported.

srikris:
We can now do the following

```
        sf1 = SFrame(range(10))
        relative_path = 'tmp/%s' % self.filename

        # Act
        pickler = gl_pickle.GLPickler(relative_path)
        pickler.dump(sf1)
        pickler.close()
        sf2 = gl_pickle.GLUnpickler(relative_path).load()
```

Previously, the '~' wasn't being expanded out to the home directory and was being saved to the current directory.


hoytak:
Looks great to me.

PR #87, opened 2015-11-11
Added SFrame.to_numpy and SArray.to_numpy functions

ylow:
It can be useful at times to have these shortcut functions
```
sframe.to_numpy()
and 
sarray.to_numpy()
```


oss_src/unity/python/sframe/test/test_sarray.py
```diff
@@ -1674,3 +1674,15 @@ def test_save_sarray(self):
         data.save(tmp_dir)
         shutil.rmtree(tmp_dir)
         print data
+
+    def test_to_numpy(self):
+        X = SArray(range(100))
```
wangping25: Is there limitation on the types that supported by SFrame but not by Numpy?

Also would be nice to have 'missing value' test too.

PR #92, opened 2015-11-19
Added an abstraction for cumulative aggregates.

srikris:
- Added a cumulative aggregate abstraction.
- Added cumulative sum (serial version).

#### Details

For a cumulative aggregate
   y <- x.cumulative_aggregate(f, w_0)

An abstraction to perform cumulative aggregates which is computed as
   y[i+1], w[i+1] = func(x[i], w[i])

 where w[i] is some arbitary state.

A user can implement cumulative_sum as follows.

```
  auto cumsum_agg = []
    (const flexible_type& v, flexible_type& y) -> flexible_type {
     y += v;
     return y
  };
  out_sa = in_sa.cumulative_aggregate(agg_fn, 0);
```

### DIFF
--- a/oss_src/unity/extensions/cumulative_aggregates.cpp
+++ b/oss_src/unity/extensions/cumulative_aggregates.cpp
@@ -78,27 +78,51 @@ void check_vector_equal_size(const gl_sarray& in) {
 
 }
 
-gl_sarray _sarray_cumulative_sum(const gl_sarray& in) {
-
+gl_sarray _sarray_cumulative_built_in_aggregate(const gl_sarray& in, 
+                                                const std::string& name) {
   flex_type_enum input_type = in.dtype();
   std::shared_ptr<group_aggregate_value> aggregator;
-  switch(input_type) {
-    case flex_type_enum::VECTOR: 
-    {
-      check_vector_equal_size(in);
-      aggregator = get_builtin_group_aggregator(std::string("__builtin__vector__sum__")); 
-      break;
+
+  // Cumulative sum, and avg support vector types.
+  if (name == "__builtin__cum_sum__") {
+    switch(input_type) {
+      case flex_type_enum::VECTOR: {
+        check_vector_equal_size(in);
+        aggregator = get_builtin_group_aggregator(std::string("__builtin__vector__sum__")); 
+        break;
+      }
+      default:
+        aggregator = get_builtin_group_aggregator(std::string("__builtin__sum__")); 
+        break;
     }
-    default:
-      aggregator = get_builtin_group_aggregator(std::string("__builtin__sum__")); 
-      break;
+  } else if (name == "__builtin__cum_avg__") {
+    switch(input_type) {
+      case flex_type_enum::VECTOR: {
+        check_vector_equal_size(in);
+        aggregator = get_builtin_group_aggregator(std::string("__builtin__vector__avg__")); 
+        break;
+      }
+      default:
+        aggregator = get_builtin_group_aggregator(std::string("__builtin__avg__")); 
+        break;
+    }
+  } else if (name == "__builtin__cum_max__") {
+      aggregator = get_builtin_group_aggregator(std::string("__builtin__max__")); 
+  } else if (name == "__builtin__cum_min__") {
+      aggregator = get_builtin_group_aggregator(std::string("__builtin__min__")); 
+  } else if (name == "__builtin__cum_var__") {
+      aggregator = get_builtin_group_aggregator(std::string("__builtin__var__")); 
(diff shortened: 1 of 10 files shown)


oss_src/unity/extensions/cumulative_aggregates.cpp
```diff
@@ -0,0 +1,101 @@
...
+  // Empty case.  
+  size_t m_size = in.size();
+  if (m_size == 0) {
+    return gl_sarray({});
+  } 
+  
+  gl_sarray_writer writer(output_type, 1);
+  AccumulatorType y = init; 
+  for (const auto& v: in.range_iterator()) {
+    writer.write(aggregate_fn(v, y), 0); // z[t+1], y[t+1] = f(x[t], y[y])
```
esamanas: Is there any error handling that needs to be done for aggregate_fn? Must the aggregate_fn not throw?

oss_src/unity/extensions/cumulative_aggregates.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <unity/lib/toolkit_function_macros.hpp>
+#include <unity/extensions/cumulative_aggregates.hpp>
+
+namespace graphlab {
+namespace cumulative_aggregates {
+
+template <typename AggregateFunctionType, 
+          typename AccumulatorType> 
+gl_sarray cumulative_aggregate(const gl_sarray& in,
+                     AggregateFunctionType aggregate_fn,
```
esamanas: It'd be cool if you added an example aggregate function.

esamanas:
Code looks good, but I thought we were implementing this within SArray and not as an extension?

srikris:
@esamanas @ylow The updated cumulative aggregates re-uses the groupby aggregate code and is built into the gl_sarray. It is implemented as a parallel prefix sum and will run in parallel on all aggregators that are associative.

```
gl_sarray _sarray_cumulative_sum(const gl_sarray& in) {

  flex_type_enum input_type = in.dtype();
  std::shared_ptr<group_aggregate_value> aggregator;
  aggregator = get_builtin_group_aggregator(std::string("__builtin__sum__")); 
  return in.cumulative_aggregate(aggregator);

}
```

oss_src/unity/extensions/cumulative_aggregates.cpp
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Throw an error if a vector is of unequal length.
+ * \param[in] gl_sarray of type vector 
+ */
+void check_vector_equal_size(const gl_sarray& in) {
+  
+  // Initialize. 
+  DASSERT_TRUE(in.dtype() == flex_type_enum::VECTOR); 
+  size_t n_threads = thread::cpu_count();
+  DASSERT_GE(n_threads, 1);
```
esamanas: I'd rather not assert in this case, as this could happen in the wild. Use std::min

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3039,3 +3039,185 @@ def sort(self, ascending=True):
...
+        expected to be of numeric type (int, float), or a numeric vector type.
+
+        Returns
+        -------
+        out : sarray[int, float, array.array]
+
+        Notes
+        -----
+         - Missing values are ignored while performing the cumulative
+           aggregate operation.
```
esamanas: Note that vectors must be same size too.

oss_src/unity/lib/gl_sarray.cpp
```diff
@@ -596,6 +597,118 @@ gl_sarray gl_sarray::subslice(flexible_type start,
...
+      for (const auto& v: this->range_iterator(start_row, end_row)) {
+        DASSERT_TRUE(thread_idx < aggregators.size());
+        if (v != FLEX_UNDEFINED) {
+          aggregators[thread_idx]->add_element_simple(v);
+        }
+      }
+    });
+
+    // Phase 2: Combine prefix-sum(s) at the end of each block.
+    for (size_t i = n_threads - 1; i > 0; i--) {
```
esamanas: Can't this be done going forwards with less steps?

esamanas:
This looks good. Address my comments and then...are we supposed to wait for testing to merge?

oss_src/unity/extensions/cumulative_aggregates.cpp
srikris: Done.

oss_src/unity/python/sframe/data_structures/sarray.py
srikris: Done.

PR #93, opened 2015-11-19
Rewrite of flexible type converter logic.

hoytak:
Rewrote the flexible type converter logic to be much more readible and
flexible.  This is the next stage of the flexible type improvements.

The logic used a different metaprogramming method documented in
flexible_type_converter.hpp.  This gives much better control over
error messages and better flexibility in adding additional types.

Also discovered and fixed some minor bugs with the gl_string classes.


oss_src/flexible_type/flexible_type_converter.hpp
```diff
@@ -13,621 +13,619 @@
...
 
-/**
- * Case 1a: Cover all the direct members of flexible_type, 
- *         EXCEPT for numeric types (flex_int, flex_float), and flex_list.
- *        - flex_string (std::string)
- *        - flex_vec (std::vector<double>)
- *        - flex_dict (std::vector<std::pair<flexible_type, flexible_type> >)
- *        - flex_date_time (boost::ptime)
+////////////////////////////////////////////////////////////////////////////////
+
```
ylow: Should prioritize exact type matches before the remaining cases since those have fast paths.

oss_src/flexible_type/flexible_type_converter.hpp
```diff
@@ -13,621 +13,619 @@
...
-    }
+template <> struct ft_converter<7> {
+  
+  template <typename T> static constexpr bool matches() {
+    typedef typename first_nested_type<T>::type  U1; 
+    typedef typename second_nested_type<T>::type U2; 
+    
+    return (is_std_pair<T>::value
+            && std::is_arithmetic<U1>::value 
+            && std::is_arithmetic<U1>::value);
```
ylow: What is this case?
should this be is_arithmetic<U2>?  Why is this case distinct from the case below?

oss_src/flexible_type/flexible_type_converter.hpp
ylow: For instance flex_list, flex_vec and flex_dict have faster conversion paths than the otherwise implied path below.

hoytak: I'm not sure that's true...  These would use the `v1.assign(v2.begin(), v2.end())` methods instead of `v1 = v2`, which should be exactly the same for identical matches of vector types.  I can add it these cases for clarity, but I think it's already as fast as possible.

oss_src/flexible_type/flexible_type_converter.hpp
hoytak: This one handles the case that the std::pair is one of numerics, so that the flexible_type type is then allowed to be a flex_vec.

haijieg:
jenkins pls

ylow:
Looks about right, but you want to add fast paths for direct matches to flexible_type types. i.e. if the passed type is exactly a flex_dict for instance, you want it to execute a single copy and no go through the longer path.

There is a template function: is_valid_flex_type<T>::value (see flexible_type_base_types.hpp) for a compile time check if T matches one of the flexible_type base types.

hoytak:
Currently, it appears this is causing issues with the variant type stuff in some cases, so I'm investigating that when I have time.

PR #107, opened 2015-12-14
Add a missing parameter to the docstring of SFrame.print_rows

rschindlauer:
SFrame.print_rows also takes a file handle that enables the user to
redirect output to something other than stdout, like a file or a string
stream. By documenting this parameter this feature becomes easier to
discover.


TobyRoseman:
The failure (lambda workers running out of memory) looks unrelated to your changes. I've restarted the build.

PR #108, opened 2015-12-15
Add rolling aggregates to SArray

esamanas:


oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3039,3 +3039,255 @@ def sort(self, ascending=True):
...
+        >>> series = pandas.Series([1,2,3,4,5])
+
+        A rolling sum with a window including the previous 2 entries including
+        the current:
+        >>> sa.rolling_sum(-2,0)
+        dtype: int 
+        Rows: 5
+        [None, None, 6, 9, 12]
+
+        Pandas equivalent:
```
srikris: I think this belongs more in the translator. I would personally prefer removing them.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3039,3 +3039,255 @@ def sort(self, ascending=True):
...
+
+    def rolling_var(self, window_start, window_end, min_observations=None):
+        if min_observations is None:
+            min_observations = (1 << 64) - 1
+        if min_observations < 0:
+            raise ValueError("min_observations must be a positive integer")
+        agg_op = '__builtin__var__'
+        return SArray(_proxy=self.__proxy__.rolling_apply(agg_op, window_start, window_end, min_observations))
+
+    def rolling_stdv(self, window_start, window_end, min_observations=None):
```
srikris: Doc strings here?

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3039,3 +3039,255 @@ def sort(self, ascending=True):
...
+            min_observations = (1 << 64) - 1
+        if min_observations < 0:
+            raise ValueError("min_observations must be a positive integer")
+        agg_op = '__builtin__min__'
+        return SArray(_proxy=self.__proxy__.rolling_apply(agg_op, window_start, window_end, min_observations))
+
+    def rolling_var(self, window_start, window_end, min_observations=None):
+        if min_observations is None:
+            min_observations = (1 << 64) - 1
+        if min_observations < 0:
```
srikris: Refactor this into a utility function. No need to copy paste this block of code so much :)

oss_src/unity/python/sframe/cython/cy_sarray.pyx
```diff
@@ -429,3 +429,9 @@ cdef class UnitySArrayProxy:
         with nogil:
             proxy = (self.thisptr.copy_range(start, step, end))
         return create_proxy_wrapper_from_existing_proxy(self._cli, proxy)
+    
+    cpdef rolling_apply(self, string fn_name, ssize_t before, ssize_t after, size_t min_observations):
```
srikris: I think you should call this builtin_rolling_apply because, in the future, we will probably have a pure rolling_apply?

oss_src/unity/lib/gl_sarray.hpp
```diff
@@ -1721,6 +1721,53 @@ class gl_sarray {
...
+   *  // Moving window encompasses 3 values behind current and current value.
+   *  auto result = a.rolling_apply(std::string("__builtin__avg__"), -3, 0);
+   *  \endcode
+   *
+   *  Produces an SArray with these values: 
+   *  \code
+   *  {NULL,NULL,NULL,1.5,2.5,3.5,4.5,5.5,6.5,7.5}
+   *  \endcode
+   */
+  gl_sarray rolling_apply(const std::string &fn_name,
```
srikris: Technically, the gl_sarray is a public interface for the SDK, so I would prefer calling this builtin_rolling_apply (just to future proof our rolling_apply in the future which might accept functions.

srikris:
Overall, this looks good to me. My only suggestion is to change the name of rolling_apply in the C++ side and the cython side to something else to future proof when we have a more generic rolling_apply.

oss_src/unity/python/sframe/data_structures/sarray.py
esamanas: I agree it's a better fit there, but my goal was to make this discoverable. From what I've heard, not a lot of people look at the translator, and I would guess most users of TimeSeries are coming from pandas.

oss_src/sframe/rolling_aggregate.cpp
```diff
@@ -0,0 +1,178 @@
...
+  return std::min(upper, std::max(lower, val));
+}
+
+// Calculate the size of an inclusive range
+size_t calculate_window_size(ssize_t start, ssize_t end) {
+  if((start < 0) && (end >= 0)) {
+    return size_t(std::abs(start) + end + 1);
+  }
+
+  auto the_size = size_t(std::abs(end - start) + 1);
```
TobyRoseman: Just curious, is there a reason you're using "auto" on the left hand side, rather than size_t?

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3040,6 +3040,701 @@ def sort(self, ascending=True):
         sf['a'] = self
         return sf.sort('a', ascending)['a']
 
+    def __check_min_observations(self, min_observations):
```
TobyRoseman: This should be static. You're not using 'self'.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3040,6 +3040,701 @@ def sort(self, ascending=True):
...
+
+        A rolling mean with a window including the previous 2 entries NOT
+        including the current:
+        >>> sa.rolling_mean(-2,-1)
+        dtype: float
+        Rows: 5
+        [None, None, 1.5, 2.5, 3.5]
+        """
+        min_observations = self.__check_min_observations(min_observations)
+        agg_op = None
```
TobyRoseman: Very minor note, but the way scoping works in Python, you don't need to declare this variable outside of the if/else block to be able to use it after the if/else block.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -3040,6 +3040,701 @@ def sort(self, ascending=True):
...
+        >>> series = pandas.Series([1,2,3,4,5])
+
+        A rolling standard deviation with a window including the previous 2
+        entries including the current:
+        >>> sa.rolling_stdv(-2,0)
+        dtype: float
+        Rows: 5
+        [None, None, 0.816496580927726, 0.816496580927726, 0.816496580927726]
+
+        Pandas equivalent:
```
TobyRoseman: I think it's really strange that we're showing Pandas equivalent code in doc strings. If people are not already familiar with pandas, this is probably going to be very confusing.

oss_src/sframe/rolling_aggregate.cpp
esamanas: No, not really. It should probably be size_t.

oss_src/unity/python/sframe/data_structures/sarray.py
esamanas: It's a fair point, but I don't think that confusion will be crippling. My thought was that most people using us are going to be used to using time series stuff in pandas since it has such good support for it and is well established, and no one really finds the translator quickly. The model that we're using for specifying the rolling window, while I think it's better than pandas, it could be difficult to figure out at first without these if you're used to specifying the absolute size of the window.

PR #109, opened 2015-12-15
Improvements to the lambda diagnostics

hoytak:
Added in lots of diagnostic information for finding pylambda worker
issues.

- Errors are printed to sys.stderr, so they appear in the console.

- If GRAPHLAB_LAMBDA_WORKER_DEBUG_MODE is set to 1 as an environment
  variable, then all log messages are pushed to stderr and detailed
  progress messages are printed to the console.

Other changes:

- The comm_server only prints out diagnostic information if the
  environment variable GRAPHLAB_COMM_SERVER_DEBUG_MODE is set
  (Otherwise, all lambda worker events and pings are dumped to the
  console.)


haijieg:
Can we remove the dependency of process from logger? Logger module should have minimal dependency as everything else seems to depend on it.

To support log with pid, we can do something like
```
global_logger().set_pid(mypid)
LOG_DEBUG_WITH_PID(...)
```

What do you think?

hoytak:
Sure -- I can do that.

PR #110, opened 2015-12-15
SFrame optimizations

ylow:
This is a deep improvement to the SFrame query execution layer. Currently, we have a rate control issue: we can only lazily execute fully linear pipelines (i.e. every operator emits output at the same rate as its inputs, and must consume every input at the same rate. For instance, a numeric operator is linear, a filtering operator is not). 

To execute non-linear pipelines requires potentially unbounded buffering at each operator node. This implements the unbounded buffer via a datastructure I am calling the "broadcast queue". 

This allows arbitrary chains of logical filter to be completely lazy. 

It became necessary to solve the following inconsistencies right now: 
```
> gl.SArray([]).sum()
None
> gl.SArray([None]).sum()
0.0
```
So we settled on the latter. i.e. 
```
> gl.SArray([]).sum() 
0.0
```

Other changes:
test_io.py. /root is not a system folder on Mac. Changed it to /System

#### Sum of an empty array returns the "0" value for the array type.
```
> gl.SArray([], float).sum() 
0.0
> gl.SArray([], int).sum() 
0
> gl.SArray([], array.array).sum() 
array.array('d')
```


ylow:
Do not merge yet. This is a deep change and can do with some more exhaustive testing especially on large datasets.

haijieg:
jenkins pls

oss_src/sframe/sframe_rows.cpp
```diff
@@ -32,16 +34,36 @@ void sframe_rows::clear() {
 
 void sframe_rows::save(oarchive& oarc) const {
   oarc << m_decoded_columns.size();
-  for (auto& i : m_decoded_columns) oarc << *i;
+  oarchive temp_inmemory_arc;
+  for (auto& i : m_decoded_columns) {
+    v2_block_impl::block_info info;
```
haijieg: Does this only apply to sframe_rows.save()/load() or shall we move it into lower level sarray column save/load?

oss_src/sframe_query_engine/util/broadcast_queue.hpp
```diff
@@ -0,0 +1,554 @@
...
+ *
+ * When data is pushed, we push directly into the center queue. But if the
+ * center queue is full, we start writing to the push_file.
+ *
+ * When data is popped, we first read from the pop-file. When there are no more
+ * pop files, we reads from the center queue. When the center queue is empty,
+ * it closes the push and swaps it to a pop file and starts reading from the pop
+ * file.
+ *
+ * Thus there can be at most one pop file.
```
haijieg: you mean at most one push file?

oss_src/sframe_query_engine/util/broadcast_queue.hpp
```diff
@@ -0,0 +1,554 @@
...
+ *  - It goes into the memory queue
+ *  - When memory queue exceeds some threshold, it gets written to disk
+ *    and a new memory queue is started. 
+ *
+ * Every consumer remembers a current file position pointer, and 
+ *  - pops from file
+ *  - or reads from the in memory queue. (a little bit of bookkeeping
+ *    is needed since the queue may get flushed to disk)  
+ *
+ * While this satisfies the property (1), this dies not satisfy property (2).
```
haijieg: dies -> does

haijieg:
LGTM. Fix unity_sarray.cxx tests and some minor doc changes.

oss_src/sframe/sframe_rows.cpp
ylow: This is just for sframe_rows. sframe_rows is normally an internal object used to emulate vector<rows> when the actual representation is vector<columns>. But on occasion it is very useful to be able to serialize and deserialize these things.

oss_src/sframe_query_engine/util/broadcast_queue.hpp
ylow: There can be at most 1 push file and 1 pop file yes.

PR #115, opened 2015-12-16
Remove unity_server separate process

haijieg:
Remove unity_server as separate c++ process and replace with a shared library with zeromq inproc communication. This change will significantly simplifies the communication layer between python client and c++ server. It also lowers the call latency by directly serializing within the same process.

The separate process architecture was proposed to support the case where client and server are running on different machines. However, it also causes slow start up time and slow I/O between client and server due to process barrier. As we learn that the use cases for the separate client/server configuration is rarely used, it is time to remove the separate unity_server process, and make it as a embed library into python process.

Apart from superficial changes to the server launch code path, the underlying communication architecture remains the same, except that we use zeromq's inproc socket instead of ipc or tcp sockets.

Change list:
- Refactor of the unity_server start code
  - system related startup code such is moved into startup_teardowns
  - unity_server related startup code including extension/toolkits/datastructure registration is moved into unity_server.cpp
  - remove original unity_server.cpp
  - unity_server_capi.cpp defines the shared library entry
- Move libjvm path search code into C++ side because the inproc server is not affected by changing LD_LIBRARY_PATH once python is started
-  remove global variable: GLOBALS_MAIN_PROCESS_BINARY
- Remove tcmalloc dependency from pylambda and rcppalmbda
- Deprecate use of "remote://" or "local://" as file protocols
- Purge unit test


oss_src/unity/python/sframe/connect/server.py
```diff
@@ -54,269 +51,92 @@ def get_logger(self):
...
         raise NotImplementedError
 
+ADDITIONAL_SUPPORT_MESSAGE = """
+There may have been an issue during installation. Please uninstall sframe
+and reinstall it, looking for errors that may occur during installation.
+"""
+
 
-class LocalServer(GraphLabServer):
+class EmbededServer(GraphLabServer):
```
ylow: Embeded is spelt Embedded

oss_src/unity/python/sframe/connect/server.py
```diff
@@ -54,269 +51,92 @@ def get_logger(self):
...
     """
-    Local GraphLab Server wraps the graphlab
-    unity_server process and implements the GraphLabServer interface.
+    Embeded Server loads unity_server into the same process as shared library.
     """
 
-    __slot__ = ["server_bin", "server_addr", "unity_log", "proc", "auth_token"]
+    SERVER_LIB = 'libunity_server.so'
```
ylow: is this always .so? How about on windows?

oss_src/unity/server/unity_server_capi.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/algorithm/string.hpp>
+#include <startup_teardown/startup_teardown.hpp>
+
+#include "unity_server.hpp"
+
+namespace graphlab {
+
+// Global embeded server and client object
+static unity_server* SERVER;
```
ylow: = nullptr;

oss_src/unity/server/unity_server_capi.cpp
```diff
@@ -0,0 +1,101 @@
...
+static cppipc::comm_client* CLIENT;
```
ylow: ditto

ylow:
+1 to merge. minor fixes.

haijieg:
@ylow Your comments are addressed

haijieg:
jenkins pls

haijieg:
jenkins pls

haijieg:
jenkins pls

haijieg:
jenkins pls

haijieg:
jenkins pls

oss_src/lambda/graph_pylambda_master.cpp
```diff
@@ -14,12 +14,24 @@ namespace graphlab {
 
 namespace lambda {
 
+static graph_pylambda_master* instance_ptr;
```
ylow: =nullptr just to be nice

haijieg:
jenkins pls

PR #116, opened 2015-12-16
restrict the number of threads used when running tests when num_procs is set

ylow:
This ought to stabilize the travis build (which runs in docker, and so we see much more CPUs than we are actually allocated and we end up using a lot of memory).


ylow:
@hoytak too

PR #119, opened 2015-12-17
Bugfixes to PyLambda Workers and process_win.cpp on Windows

hoytak:
This PR includes a number of fixes to the PyLambda worker code that
could cause issues on windows.

- Improvements to process spawning on windows.  Several potential bugs
  were fixed that could cause issues.  In addition, stderr now
  redirects correctly.

- Made winsock2.hpp include consistent.  We need to include the
  winsock2.h header before windows.h, as using the outdated winsock.h
  header can cause issues when linking against libraries that use the
  new winsock2.h header such as the mongoose metric reporting.  The
  winsock2.h was introduced in Windows 95, but windows.h still
  includes winsock.h by default.  From the internet, it appears that
  mixing them can cause strange issues and crashes.

- Updated pylambda_worker logic so logs can be dumped to a file.

  If the GRAPHLAB_LAMBDA_WORKER_LOG_FILE environment variable is set
  to a non-empty string, it is assumed to be an open file and the
  global logger logs to that file.

  If GRAPHLAB_LAMBDA_WORKER_DEBUG_MODE, then all DEBUG items are
  logged to this file; if this is set but the above is not, then
  everything is logged to the console.  If it is not set but the above
  is, then INFO logs and above are logged to the file.

  If neither variable is set, then only ERROR and above are logged to
  the console.

- Added post-spawn diagnostic routines to windows process
  spawning. Now, the process spawning routing waits 100ms or so, then
  queries the process status, logging return codes if they fail.

- Added SetDLL resolution to spawned worker process.  Now the
  _pylambda_worker.py script also does the dll resolution using the
  kernel32 routines.

- Strengthened error reporting code in _pylambda_worker.py.  The error
  reporting routines in _pylambda_worker.py now respect the
  GRAPHLAB_LAMBDA_WORKER_DEBUG_MODE and
  GRAPHLAB_LAMBDA_WORKER_LOG_FILE environment variables.

- Moved pylambda worker to isolated scripts directory.  It appears
  that the pylambda workers could fail because, with
  _pylambda_worker.py in the root directory, that path was prefered
  for importing graphlab/sframe and this could cause relative import
  errors in the boost.python module.  Moving it to
  _scripts/_pylambda_worker.py fixes this problem.

Change-Id: If5257cdff5ac6b28b59164d705779ce65777bf2f


hoytak:
@ylow @esamanas -- I hope this clears up all the issues, or clearly exposes what they are.  I hit a bunch of intermittent errors going through things, which are fixed as part of this PR.  It currently works on windows (unless I forgot a commit or something).

oss_src/lambda/pylambda_worker.cpp
```diff
@@ -32,25 +32,47 @@ int _pylambda_worker_main(const char* _root_path, const char* _server_address) {
...
+      global_logger().set_log_to_console(true, true);
+    }
   } else {
     global_logger().set_log_level(LOG_INFO);
+    if(!log_file_string.empty()) {
+      // Log errors to console
+      global_logger().set_log_to_console(true, true);
+      global_logger().set_log_file(log_file_string);
+      LOG_DEBUG_WITH_PID("Logging lambda worker logs to " << log_file_string);
+      global_logger().set_log_to_console(false);
```
esamanas: I think you should be able to find a way to not repeat yourself here.

oss_src/logger/logger.hpp
```diff
@@ -558,6 +572,8 @@ class file_logger{
   bool log_to_stderr;
   int log_level;
 
+  size_t reference_pid = 0;
+
```
esamanas: This should maybe start as size_t(-1), since 0 can be a valid (though unlikely) PID.

oss_src/unity/python/sframe/_scripts/_pylambda_worker.py
```diff
@@ -0,0 +1,136 @@
...
+        _write_out_file_name = _write_out_file_name + "-init"
+        _write_log("Logging initialization routines to %s." % _write_out_file_name)
+        try:
+            _write_out_file = open(_write_out_file_name, "w")
+        except Exception, e:
+            _write_log("Error opening '%s' for write: %s" % (_write_out_file_name, repr(e)))
+            _write_out_file = None
+
+    if debug_mode:
+        print "PyLambda script called with no IPC information; entering diagnostic mode."
```
esamanas: I don't think this is completely true anymore, correct?

oss_src/process/process_win.cpp
```diff
@@ -39,32 +39,50 @@ bool process::launch(const std::string &cmd,
...
-      CREATE_NO_WINDOW,
-      NULL,
-      NULL,
-      &startup_info,
-      &proc_info);
+      c_arglist, // command line
+      NULL,      // process security attributes
+      NULL,      // primary thread security attributes
+      TRUE,      // handles are inherited
+      0,         // creation flags
```
esamanas: Is there a reason CREATE_NO_WINDOW was removed here?

oss_src/process/process_win.cpp
```diff
@@ -39,32 +39,50 @@ bool process::launch(const std::string &cmd,
...
+
   BOOL ret;
+
+  ret = DuplicateHandle( GetCurrentProcess(), GetStdHandle(STD_ERROR_HANDLE),
+                         GetCurrentProcess(), &m_stderr_handle, 0, TRUE, DUPLICATE_SAME_ACCESS);
+
+  if(!ret) {
+    auto err = GetLastError();
+    logstream(LOG_WARNING) << "Failed to create duplicate stderr file handle: " << get_last_err_str(err) << std::endl;
+    logstream(LOG_WARNING) << "  Attempting to continue with : " << get_last_err_str(err) << std::endl;
```
esamanas: Is this the intended error message? It seems like it will return the same error string twice.

oss_src/process/process_win.cpp
hoytak: Yes, it turns out that if it is on, then stderr and stdout aren't going to be forwarded. http://stackoverflow.com/questions/19051769/createprocess-redirect-child-stdout-to-parent-stdout-in-c-windows

oss_src/logger/logger.hpp
hoytak: okay, could do.

oss_src/unity/python/sframe/_scripts/_pylambda_worker.py
hoytak: It is within the _pylambda_worker.py file, but not the pylambda_worker.cpp file.  they mean slightly different things... though I could update the variable name to reflect that.

oss_src/lambda/pylambda_worker.cpp
hoytak: I tried, but it's slightly different , so it seemed easier to just copy paste here.

esamanas: Keep in mind that the logic of what log messages go where depending on the settings of two environment variables is already very non-trivial to follow. I hoped the code would make it clearer, but I was confronted with this, which is ridiculous. This needs to be clear to anyone that that looks at this later, including you, because you'll forget what you meant or that you did this because "it seemed easier".

oss_src/unity/python/sframe/_scripts/_pylambda_worker.py
esamanas: Okay, looking at it again, I see that now. It is pretty confusing to have a debug_mode toggled based on the amount of arguments, and then a different debug mode toggled by an environment variable ending in "DEBUG_MODE". Perhaps you could call the one that runs with no IPC a "dry run"?

oss_src/fileio/temp_files.cpp
```diff
@@ -9,7 +9,7 @@
 #ifndef _WIN32
 #include <pwd.h>
 #else
-#include <windows.h>
+#include <process/gl_windows.hpp>
```
esamanas: Also, why is the windows.h wrapper in "process"?

oss_src/process/process_win.cpp
ylow: But without CREATE_NO_WINDOW, under certain conditions, launching of pylambda_worker will result in NCPUs worth of cmd window popups.

oss_src/lambda/pylambda_worker.cpp
hoytak: Okay, I'll add a comment to clear that up.  I think I see where the confusion is.

oss_src/process/process_win.cpp
hoytak: Okay, well, I will attempt to look for other options here, then.  Yikes.  That's annoying.

oss_src/unity/python/sframe/_scripts/_pylambda_worker.py
hoytak: Great suggestion.  I'll try to do that.

oss_src/process/process_win.cpp
ylow: If we have a file output mechanism already, do we still need stderr/stdout forwarding? Can we eliminate that entirely as an alternative?

esamanas: @ylow Do you remember what the case was that we noticed that caused CREATE_NO_WINDOW to be in there? I seem to recall it was more of a thing that we added after reading MSDN docs and thinking "yeah, we should probably have that". 

If we were to completely forgo stderr forwarding (which might be good, as it's not very discoverable output when using IPython Notebook), I think I would want pylambda workers to push their stderr to a file at all times then, not just when the variable is set. That way the error message is at least attainable for an error that is not reproducible.

hoytak: I like that option.  Along with that, I think it should be passed as an argument from the server through to the spawning process.  This would eliminate the need for those very annoying environment variables.

oss_src/process/process_win.cpp
```diff
@@ -75,6 +110,34 @@ bool process::launch(const std::string &cmd,
...
 
+  // Wait up to 100 milliseconds before querying the process for it's status.
+  DWORD dwMillisec = 100;
+  DWORD dwWaitStatus = WaitForSingleObject(m_proc_handle, dwMillisec );
+
+  if(dwWaitStatus == WAIT_FAILED) {
+    auto err = GetLastError();
+    logstream(LOG_WARNING) << "Error in WaitForSingleObject after CreateProcess: "
+                           << get_last_err_str(err) << std::endl;
+  }
```
ylow: Why wait up to 100ms?

hoytak: It seemed 100ms caught the errors I was seeing in python not starting or library import errors, which seemed to be most of the problems.  Balancing this is that we don't want to wait so long that the user notices it.  This seemed like a good compromise.

PR #121, opened 2015-12-21
Change engine start error message for SFrame.

znation:
* There should not be a reference to GraphLab Create (this is the SFrame engine).
* There should not be a reference to Dato support (instead, we should direct users to GitHub since this is an open source project).


znation:
@timmuss Please review as well, thanks!

ylow:
Will merge, but this will be deprecated by #115 .

PR #122, opened 2015-12-22
Locked version number to 1.8

ylow:
Switching to using hard-coded version numbers.

Key changes to build scripts:

This should not be used
  ./make_egg.sh --version=1.8

instead
  ./make_egg.sh --build_number=123
or
  ./make_egg.sh --build_number=123.gpu

Note that build_number is a string.

To make porting incremental, for now, ./make_egg.sh --version=BLAH is
*identical* in behavior to ./make_egg.sh --build_number=BLAH Hence the
outcome of using old un-updated build processes is simply that the
version number will be the hard-coded version number, and the build
number is the value specified in make_egg.sh

A script

  ./update_version_number.sh

can be used to update all the hardcoded version numbers. The way it
works is by tagging all the locations with the version strings.
For instance, in Python,

        version_number='1.8'#{{VERSION_STRING}}

we use a regex to find the pattern '.*'#{{VERSION_STRING}} and replace
the version number

similarly, in C++ we look for the pattern.

        version_number="1.8"//#{{VERSION_STRING}}

While making the version number system saner, the metrics system is
also cleaned up somewhat, removing librato.


oss_local_scripts/run_cpp_tests.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
+
+import os
+import time
+import pickle
+import random
```
TobyRoseman: This isn't getting used.

TobyRoseman:
Looks good to me.

PR #125, opened 2016-01-04
Add support for numpy and pandas datetime objects

esamanas:
This adds support for numpy.datetime64 and pandas.Timestamp.

Conversion from numpy.datetime64 results in a datetime.datetime with timezone set explicitly to UTC, since numpy stores its info in UTC and does not keep track of timezones.

Conversion from pandas.Timestamp uses that type's own "to_datetime" method.


dbickson:
Another issue which is slightly related: http://forum.dato.com/discussion/1609/create-sframe-from-feature-matrix-x-and-target-vector-y

PR #131, opened 2016-01-12
Fix for a memory leak when creating empty sframe or sarray

ylow:
For for a memory leak when creating empty SFrame (no rows or no columns) and empty sarrays (no rows).

In addition, as an optimization we maintain one empty SArray and one empty SFrame. But there is an annoying termination race issue between 
 - clearing of the cache files
 - deletion of the empty SArray/SFrame

A proper solution is non-trivial. At the moment the cleanest solution is simply to leak the empty SArray / empty SFrame.


wangping25:
Looks good to me!

PR #132, opened 2016-01-14
Complete removal of boost::python.

hoytak:
This pull request finishes the support for running the pylambda workers in cython and removes the use boost::python from the code base.  

- cy_pylambda_worker.pyx now does the heavy lifting for lambda evaluation.
- centralized python callback function for exception handling.
- spark unity now a python script that enables cython callbacks.


haijieg:
jenkins pls

ylow:
@esoroush Can you test this?

hoytak:
The travis build failure is odd -- it's compiling an old version of the file that no longer has that issue.  Commit cd2ec63 doesn't have that in there.  Is there a way to retrigger travis?

haijieg:
jenkins pls

oss_local_scripts/make_egg.sh
```diff
@@ -103,6 +103,9 @@ build_source() {
   echo -e "\n\n\n================= Build ${BUILD_NUMBER} ================\n\n\n"
   # Configure
   cd ${WORKSPACE}
+  # Clean up, except for the deps.
+  rm -rf "${WORKSPACE}/debug" "${WORKSPACE}/release"
```
ylow: Eh... Lets not do this. In almost all circumstances this is unnecessary. And it slows down the build quite significantly since now it *really* has to rebuild everything.

hoytak: True, although the travis logs show ccache is activated, so it shouldn't be too bad there.

TobyRoseman:
@hoytak - I think this PR could also remove a fair amount of stuff `override_cmake/ExternalProjectBoost.cmake` too.

oss_src/unity/python/sframe/cython/cy_pylambda_workers.pyx
```diff
@@ -0,0 +1,473 @@
...
+        cdef list bad_values
+
+        for j, k in enumerate(mut_keys):
+
+            try:
+                src_val = ret_dict[k]
+                src_key_count += 1
+                
+            except KeyError:                
+                src_val = ref_dict[k]
```
TobyRoseman: I don't understand this line. There's only two lines in the try block. One can not possibly generate a KeyError. The other line is just copied in the except block. Why?

hoytak: It's not actually copied -- ref vs. ret.    One is the reference block, the other is the returned block from the user.

TobyRoseman: Thanks. That makes a lot more sense.

hoytak:
@ylow @emad -- are we ready to merge this?

hoytak:
OMG this makes me so happy to finally have this in.

PR #137, opened 2016-01-19
Remove some unused toolkits utils.

papayawarrior:
The deleted utils are used by GraphLab Create, but not used
anywhere in the SFrame package.


papayawarrior:
Aborted the build, I'm going to put some of the functions back in for now because it would be too much work to update all of the calling modules.

papayawarrior:
Update finished. Ready to review.

PR #139, opened 2016-01-22
Rewrite of cy_variant.pyx for faster variant type conversion

hoytak:
This PR gives cy_variant.pyx the same love that cy_flexible_type.pyx
has received.

### DIFF
--- a/oss_src/unity/python/sframe/cython/cy_variant.pyx
+++ b/oss_src/unity/python/sframe/cython/cy_variant.pyx
@@ -6,6 +6,8 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 '''
 # cython: c_string_type=str, c_string_encoding=UTF-8
+# cython: boundscheck=False
+# cython: wraparound=False
 # distutils: language = c++
 import cython
 import types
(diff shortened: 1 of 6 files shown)


oss_src/unity/python/sframe/cython/cy_variant.pyx
```diff
@@ -33,170 +36,593 @@ from .cy_unity cimport is_function_closure_info
...
+    # First, check if all the keys are string values.  If they are, we
+    # are doing a variant_map_type.  Otherwise, encode it as a
+    # flexible type.
+
+    cdef bint output_is_varmap = True
+
+    for k in d.iterkeys():
+        if type(k) is not str:
+            output_is_varmap = False
+            break
```
ylow: I don't understand the logic here. If all keys are strings, it still can be a flexible_type via flex_dict.
The oddity is when the values contain non-trivial types like SArray/SFrame/etc.

hoytak: Does flex_dict always have priority over varmap?   If that is the case, then I'll update this logic.

ylow: May have to do something similar like the list case above: just repeated on keys() and values() possibly.

hoytak: Can do.

ylow: yeah. Generally, if we can contain something in a flexible_type, we should: it is the simpler type.

PR #141, opened 2016-01-25
fix a memory leak in lambda worker

haijieg:
Fix a bug that the captured lambda function was leaked and not released properly.


hoytak:
Good catch.

PR #147, opened 2016-01-27
Add DBAPI2 support for SFrame 

esamanas:
This results in two top-level functions added to SFrame: from_sql and
to_sql. They take a DBAPI2 connection object and can convert database
results to an SFrame and write SFrame data to a database table.

This patch also includes SArrayBuilder and SFrameBuilder, which are
incremental interfaces for building SArrays and SFrames. from_sql uses
SFrameBuilder internally.


oss_src/unity/python/sframe/data_structures/sframe_builder.py
```diff
@@ -0,0 +1,181 @@
...
+    | X1 |  X2 | X3 |
+    +----+-----+----+
+    | 1  | 1.0 | 1  |
+    | 2  | 2.0 | 2  |
+    | 3  | 3.0 | 3  |
+    +----+-----+----+
+    [3 rows x 3 columns]
+
+    """
+    def __init__(self, column_types, column_names=None, num_segments=1, history_size=10):
```
ylow: Might be useful to have an output location as an optional parameter.

oss_src/unity/python/sframe/cython/cy_sarray_builder.pxd
```diff
@@ -0,0 +1,43 @@
+'''
```
ylow: Feel slightly mixed about this. Might prefer this to go via the SDK interface rather than the cython interface. On the other hand the cython interface is probably still going to be somewhat faster.

oss_src/unity/extensions/sarray_builder.cpp
```diff
@@ -0,0 +1,115 @@
...
+  for(const auto &i : vals) {
+    this->append(i, segment);
+  }
+}
+
+flex_type_enum unity_sarray_builder::get_type() {
+  return m_ary_type;
+}
+
+std::vector<flexible_type> unity_sarray_builder::read_history(size_t num_elems) {
```
ylow: What is read_history for?

esamanas: It's more a slight help for debugging or using it interactively. I thought if someone has a loop going where they're adding values and something errors out, this could let them know where they were in their data.

oss_src/unity/python/sframe/cython/cy_sarray_builder.pxd
esamanas: I actually started in the SDK, but after talking with Hoyt about taking a list of types as a parameter to several functions, it seemed like cython was the best route.

oss_src/unity/extensions/sarray_builder.cpp
```diff
@@ -0,0 +1,115 @@
...
+        m_ary_type = get_common_type(m_types_inserted);
+      } catch(std::string &e) {
+        m_types_inserted.erase(ins_ret.first);
+        log_and_throw(std::string("Append failed: ") +
+            flex_type_enum_to_name(in_type) + std::string(" type is "
+              "incompatible with types of existing values in this SArray."));
+      }
+    }
+  }
+
```
ylow: So, doing "online" type inference like this is slightly problematic. By the time appends happen you should lock down the type. This "might" work sometimes but is not really guaranteed.

oss_src/unity/extensions/sarray_builder.cpp
```diff
@@ -0,0 +1,115 @@
...
+    log_and_throw("History is invalid when closed.");
+
+  if(num_elems > m_history->size())
+    num_elems = m_history->size();
+
+  if(num_elems == size_t(-1))
+    num_elems = m_history->size();
+
+  std::vector<flexible_type> ret_vec(num_elems);
+  auto riter = m_history->rbegin();
```
ylow: simpler
```
std::copy(m_history->rbegin(), m_history->r_end(), ret_vec.begin())
```

oss_src/unity/extensions/sarray_builder.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <boost/circular_buffer.hpp>
+#include <unity/lib/api/unity_sarray_builder_interface.hpp>
+
+namespace graphlab {
+
+// forward declarations
+template <typename T>
+class sarray;
+
+class unity_sarray_builder: public unity_sarray_builder_base {
```
ylow: Lets add some doxygen docs here.

oss_src/unity/extensions/sarray_builder.hpp
```diff
@@ -0,0 +1,48 @@
...
+ public:
+  void init(size_t num_segments, size_t history_size, flex_type_enum dtype);
+
+  void append(const flexible_type &val, size_t segment);
+  void append_multiple(const std::vector<flexible_type> &vals, size_t segment);
+
+  flex_type_enum get_type();
+  std::vector<flexible_type> read_history(size_t num_elems=size_t(-1));
+  std::shared_ptr<unity_sarray_base> close();
```
ylow:
```
unity_sarray_builder(const unity_sarray_builder&) = delete
unity_sarray_builder& operator=(const unity_sarray_builder&) = delete
```

oss_src/unity/lib/unity_sarray_builder.cpp
```diff
@@ -0,0 +1,118 @@
+/**
+ * Copyright (C) 2015 Dato, Inc.
+ * All rights reserved.
```
ylow: There appear to be 2 copies of this file. sarray_builder.cpp (above in unity/extensions) and this file?

oss_src/unity/lib/unity_sarray_builder.hpp
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (C) 2015 Dato, Inc.
+ * All rights reserved.
```
ylow: ditto duplicated file

oss_src/unity/lib/unity_sframe_builder.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <sframe/sframe.hpp>
+#include <boost/circular_buffer.hpp>
+#include <unity/lib/api/unity_sframe_builder_interface.hpp>
+
+typedef boost::circular_buffer<std::vector<graphlab::flexible_type>> row_history_t;
+
+namespace graphlab {
+
+class unity_sframe_builder: public unity_sframe_builder_base {
+ public:
```
ylow: similar to unity_sarray_builder
doxygen docs. + deletion of copy constructor and copy assignment

oss_src/unity/python/sframe/data_structures/sarray_builder.py
```diff
@@ -0,0 +1,147 @@
...
+        if not hasattr(data, '__iter__'):
+            raise TypeError("append_multiple must be passed an iterable object")
+        tmp_list = []
+        block_pos = 0
+        first = True
+        while block_pos == self._block_size or first:
+            first = False
+            for i in data:
+                tmp_list.append(i)
+                ++block_pos
```
ylow: ++?

oss_src/unity/python/sframe/data_structures/sframe_builder.py
```diff
@@ -0,0 +1,181 @@
...
+        if not hasattr(data, '__iter__'):
+            raise TypeError("append_multiple must be passed an iterable object")
+        tmp_list = []
+        block_pos = 0
+        first = True
+        while block_pos == self._block_size or first:
+            first = False
+            for i in data:
+                tmp_list.append(i)
+                ++block_pos
```
ylow: ++?

oss_src/unity/python/sframe/data_structures/sframe_builder.py
```diff
@@ -0,0 +1,181 @@
...
+            sequentially, starting with 0. Any value in segment 1 will be after
+            any value in segment 0, and the order of rows in each segment is
+            preserved as they are added.
+        """
+        if not hasattr(data, '__iter__'):
+            raise TypeError("append_multiple must be passed an iterable object")
+        tmp_list = []
+        block_pos = 0
+        first = True
+        while block_pos == self._block_size or first:
```
ylow: Also, see comment above in SarrayBuilder. This code is awkward

oss_src/unity/python/sframe/data_structures/sarray_builder.py
```diff
@@ -0,0 +1,147 @@
...
+            sequentially, starting with 0. Any value in segment 1 will be after
+            any value in segment 0, and the order of elements in each segment is
+            preserved as they are added.
+        """
+        if not hasattr(data, '__iter__'):
+            raise TypeError("append_multiple must be passed an iterable object")
+        tmp_list = []
+        block_pos = 0
+        first = True
+        while block_pos == self._block_size or first:
```
ylow: This code seems really awkward. I am not sure if it should work at all.
Since the loop over data is on the inside, it seems like it will just continuously insert the first block_size rows.

Cleaner will be
```
buffer = []
for i in data:
  insert into buffer
  if buffer is full:
    flush buffer
    buffer = []
if buffer is not empty:
  flush buffer
```

oss_src/unity/python/sframe/data_structures/sframe_builder.py
esamanas: So, in that case, close would still return the SFrame, it's just that the SFrame gets saved in the specified location for later?

ylow: Yup.

oss_src/unity/python/sframe/data_structures/sarray_builder.py
esamanas: It is awkward code. Your way is cleaner. For whatever reason, at the time it felt like the best way. I can change it.

It does work though. Once the block is full, the code breaks from the inside loop.

esamanas: Just kidding it only worked because of my stupid C++ style increment and it was caching everything. Ha!

oss_src/unity/python/sframe/data_structures/sframe_builder.py
esamanas: I'll probably still add this, I just didn't in this batch of changes.

oss_src/unity/lib/unity_sarray_builder.cpp
```diff
@@ -0,0 +1,114 @@
...
+  if(segment >= m_out_iters.size()) {
+    log_and_throw("Invalid segment number!");
+  }
+
+  m_history->push_back(val);
+  auto in_type = val.get_type();
+  // Only care about types if a dtype was not provided
+  if(m_given_dtype == flex_type_enum::UNDEFINED &&
+      in_type != flex_type_enum::UNDEFINED) {
+    auto ins_ret = m_types_inserted.insert(in_type);
```
ylow: If only 1 type is allowed, we don't need m_types_inserted.
Just need to compare in_type vs m_ary_type.

ylow:
Overall looks good. One other minor change. And we could fix up the save thing in a future PR.

oss_src/unity/lib/unity_sarray_builder.cpp
esamanas: yessir, 'twas a hack of laziness. I'll fix it.

PR #156, opened 2016-02-01
Python3 + boost::python removal

hoytak:
This PR aggregates the work on python 3 support by @TobyRoseman and myself.  As part of this process, it pulls in the boost::python removal change and numerous small changes to get python 3 to work.


ylow:
Ok to merge I think. We will need to remove the override_cmake and put up a new dato-deps potentially.

Issue #159, opened 2016-02-03
HDFS Sometimes Fails on Mac when Hadoop is compiled with newer JDK (1.8)


ylow:
When Hadoop is compiled with a new JVM (1.7/1.8), HDFS paths no longer work failing with 
```
java.lang.UnsupportedClassVersionError: org/apache/hadoop/fs/FileSystem : Unsupported major.minor version 51.0
	at java.lang.ClassLoader.defineClass1(Native Method)
...
```

The reason is slightly complex.

Our prebuild libhdfs.so links against the old location of the JVM (/System/Library/Frameworks/JavaVM.framework/Versions/). 
```
> otool -L libhdfs.so
libhdfs.so:
	/Users/sethu/Downloads/hadoop-2.3.0-src/hadoop-hdfs-project/hadoop-hdfs/src/target/usr/local/lib/libhdfs.0.0.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/JavaVM.framework/Versions/A/JavaVM (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```
This is also why when HDFS is used on OS X it asks to download the legacy version of Java from Apple. The JVM in this location is always 1.6.*. However, if Hadoop is compiled with a newer JDK this will fail since JVM 1.6.* will not load classes compiled with a newer JDK (1.8).

All newer JVM versions are installed in /Library/Java/JavaVirtualMachines/jdk..../. And this path is discoverable by running /usr/libexec/java_home. So the following script fixes the issue by using install_name_tool to redirect libhdfs.so linking to go to the new jvm instead
```
#!/bin/bash

glc_path=`python -c "import graphlab; print graphlab.__file__"`
libhdfs_path=$(dirname "${glc_path}")/libhdfs.so
newlibjvm_path=`/usr/libexec/java_home`/jre/lib/server/libjvm.dylib

install_name_tool -change /System/Library/Frameworks/JavaVM.framework/Versions/A/JavaVM $newlibjvm_path $libhdfs_path
```

However, our use of a wheel distribution means that we don't get the opportunity to run arbitrary code on installation to fix this path. The solution is to preload the correct libjvm.dylib on a Mac. The following works too
```
DYLD_INSERT_LIBRARIES=`/usr/libexec/java_home`/jre/lib/server/libjvm.dylib python
```
[...]

PR #160, opened 2016-02-03
Fix read/write hdfs

haijieg:
Use more robust libjvm.dylib for libhdfs on OS X

Inject `/usr/libexec/java_home`/jre/lib/server/libjvm.dylib to
be the first search location for libjvm. This is the most robust
way to get the latest system libjvm.

Also fix another bug in hdfs protocol validation

This should resolve issue #159

### DIFF
--- a/oss_src/fileio/libhdfs_shim.cpp
+++ b/oss_src/fileio/libhdfs_shim.cpp
@@ -384,6 +384,26 @@ extern  "C" {
     search_prefixes = {""};
     search_suffixes = {""};
     file_name = "libjvm.dylib";
+
+    // Run /usr/libexec/java_home to get the libjvm location
+    std::string libjvm_location = "";
+    std::string java_home_cmd = "/usr/libexec/java_home";
+    try {
+      process p;
+      p.popen(java_home_cmd,
+              std::vector<std::string>(),
+              STDOUT_FILENO);
+      libjvm_location = p.read_from_child();
+      logstream(LOG_INFO) << "Obtain JAVA_HOME from " << java_home_cmd << ": " << libjvm_location << std::endl;
+    } catch (...) {
+      logstream(LOG_WARNING) << "Error running " << java_home_cmd << std::endl;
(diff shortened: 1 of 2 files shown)


oss_src/fileio/libhdfs_shim.cpp
```diff
@@ -384,6 +385,26 @@ extern  "C" {
...
+              STDOUT_FILENO);
+      libjvm_location = p.read_from_child();
+      logstream(LOG_INFO) << "Obtain JAVA_HOME from " << java_home_cmd << ": " << libjvm_location << std::endl;
+    } catch (...) {
+      logstream(LOG_WARNING) << "Error running " << java_home_cmd << std::endl;
+      libjvm_location = "";
+    }
+    if (!libjvm_location.empty()) {
+      // Make this location to be searched first `/usr/libexec/java_home`/jre/lib/server/libjvm.dylib
+      search_prefixes.insert(search_prefixes.begin(), libjvm_location);
```
ylow: ah. the bug report was ambiguous. it is not " `/usr/libexec/java_home`" or "jre/lib/server/libjvm.dylib".
where x = `/usr/libexec/java_home`
the path is "$x/jre/lib/server/libjvm.dylib"

See the shell script in the bug report

haijieg: I believe this does the right thing.

The output of /usr/libexec/java_home is used as `search_prefix`.
'/jre/lib/server' is used as `search_suffix`.
'libjvm.dylib' is the `filen_ame`.

At the end of the function, the potential search paths are generated by
doing combinations of search_prefix / search_suffix / file_name

__

<http://www.dato.com/>   Jay (Haijie) Gu
CoFounder & Principal Data Scientist
jgu@dato.com <name@dato.com>   <https://twitter.com/datoinc>

ylow: Ah. I see it now. ok. that's right then.

PR #163, opened 2016-02-04
#160 is incomplete

ylow:
because /usr/libexec/java_home returns a string with a "\n" at the end. It is necessary to strip whitespace.


ylow:
You test this by running

>>> DYLD_PRINT_LIBRARIES=1 python

and run the following script

```
import sframe
sframe.SFrame.read_csv('hdfs://anywhere')
```
And see if /Library/Java/JavaVirtualMachines/.../libjvm.dylib is loaded

PR #172, opened 2016-02-11
Added 2 new groupby aggregates.

srikris:
- FREQ_COUNT (Frequency counter for strings/ints)
   sf.groupby("user",
         {'rating_distinct':gl.aggregate.FREQ_COUNT('rating')})

- DISTINCT (distinct values in each group)
   sf.groupby("user",
         {'rating_distinct':gl.aggregate.DISTINCT('rating')})

### DIFF
--- a/oss_src/sframe/groupby_aggregate_operators.hpp
+++ b/oss_src/sframe/groupby_aggregate_operators.hpp
@@ -979,7 +979,8 @@ class quantile : public group_aggregate_value {
 
 
 /**
- * Implements an aggregator that convert two values from two column into a key/value
+ * Implements an aggregator that convert two values from two column 
+ * into a key/value
  * value inside a dictionary
  */
 class zip_dict : public group_aggregate_value {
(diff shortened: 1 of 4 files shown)


oss_src/sframe/groupby_aggregate_operators.hpp
```diff
@@ -1255,9 +1256,107 @@ class count_distinct: public group_aggregate_value {
...
+  group_aggregate_value* new_instance() const {
+    frequency_count* ret = new frequency_count;
+    return ret;
+  }
+
+  void add_element_simple(const flexible_type& flex) {
+    if (m_values.find(flex) == m_values.end()) {
+      m_values[flex] = 1;
+    } else {
+      m_values[flex] += 1;
```
ylow: This ends up performing 2 lookups. once for find() and once for the map indexing.

Since the map value is guaranteed to zero-initialize when value type is an integer, all this code has to do is only:
```
m_values[flex] += 1;
```

srikris: Fixed these comments.

PR #177, opened 2016-02-16
Port changes from closed source to SFrame

znation:


TobyRoseman:
Discussed in person. We agree this is not the correct approach.

PR #182, opened 2016-02-18
Fix incorrect cache eviction

haijieg:
Fix the bug where the fileio cache layer allows read of a
evicted cache_block.

The cache eviction relies on the evicting thread being the only
holder of the cache_block. However, the union_fstream took an optimization
to directly grab the underlying input stream of a cache_stream,
thus giving up the cahce_block handle, allowing eviction to happen.

The fix is to remember the orignal cachestream which holds reference to
the cache_block. This prevents the block from being evicted.


haijieg:
Build failed on python3.4 due to unrelated error in the change in groupby_aggregate test. OK to merge?

PR #183, opened 2016-02-18
[#Grouped SFrame] Added some enhancements to GroupedSFrame.

srikris:
- GroupedSFrame now has additional methods to access underlying data.
- GroupedSFrame also has a `group_info` method that gives you access
  to the groups and sizes of each group quickly.


esamanas:
Looks good

PR #187, opened 2016-02-19
Fix memory leak of triple_apply

haijieg:
class batch_triple_apply_visitor needs a virtual destructor,
otherwise, class lambda_triple_apply_visitor which inherits
from it will not be properly cleared, and leaks memory.


haijieg:
Fix https://github.com/dato-code/SFrame/issues/185

PR #196, opened 2016-02-23
Modified the code to generate the SDK.

srikris:
- The SFrame SDK example does not work because a file in process was
missing. This file is now added to the sdk-generation.


TobyRoseman:
Thanks for fixing this.

PR #210, opened 2016-03-04
Fix ephemeral crash/hang on termination due to sigchild handler

haijieg:
The sigchld handler for reaping child process relies on a global
std::set which does process id bookkeeping. The global variable
could be destroyed before the the signal handler is called, causing
undefined behavior on termination.

### DIFF
--- a/oss_src/process/process_unix.cpp
+++ b/oss_src/process/process_unix.cpp
@@ -16,7 +16,6 @@
 #include <logger/logger.hpp>
 #include <boost/filesystem.hpp>
 #include <parallel/mutex.hpp>
-#include <pthread.h>
 
 namespace fs = boost::filesystem;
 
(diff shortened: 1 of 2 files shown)


oss_src/process/process_unix.cpp
```diff
@@ -32,17 +33,21 @@ namespace fs = boost::filesystem;
  * handler.
  */
 graphlab::mutex sigchld_handler_lock;
-std::set<size_t> proc_ids_to_reap;
+
```
ylow: prefer std::call_once rather than pthread_once. It is somewhat more portable.
See oss_src/fileio/block_cache.cpp for an example

haijieg: fixed

oss_src/process/process_unix.cpp
```diff
@@ -32,17 +32,27 @@ namespace fs = boost::filesystem;
...
 static void sigchld_handler(int sig) {
   // loop through the set of stuff to reap
   // and try to reap them
-  auto iter = proc_ids_to_reap.begin();
-  while (iter != proc_ids_to_reap.end()) {
-    auto cur_iter = iter;
-    ++iter;
-    if (waitpid(*cur_iter, NULL, WNOHANG) > 0) {
-      proc_ids_to_reap.erase(cur_iter);
+  auto iter = __proc_ids_to_reap->begin();
```
ylow: should compare against NULL or use get_proc_ids_to_reap() instead of getting to the pointer directly. If get_proc_ids_to_reap() is not called prior to a SIGCHLD, this will be a null pointer access.

PR #215, opened 2016-03-08
Add SFrame recognition for two types

esamanas:
Types are decimal.Decimal and datetime.timedelta, both converted to
float. The timedelta type will represent the number of seconds in the
interval.

Also handles the case of the NaT (Not a Time) type in pandas and numpy.

Addresses https://github.com/dato-code/documents/issues/3717
and https://github.com/dato-code/documents/issues/3755

### DIFF
--- a/oss_src/unity/python/sframe/cython/cy_flexible_type.pyx
+++ b/oss_src/unity/python/sframe/cython/cy_flexible_type.pyx
@@ -142,6 +142,7 @@ import datetime
 import calendar
 import collections
 import types
+import decimal
 
 from libc.stdint cimport int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t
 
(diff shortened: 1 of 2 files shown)


oss_src/unity/python/sframe/cython/cy_flexible_type.pyx
```diff
@@ -1223,8 +1230,9 @@ cdef inline tr_datetime64_to_ft(flexible_type& ret, v):
     # Since flexible type datetime only goes down to microseconds, convert to
     # this. If higher resolution, this will truncate values
     cdef object as_py_datetime = v.astype('M8[us]').astype('O')
-    as_py_datetime = as_py_datetime.replace(tzinfo=GMT(0))
-    tr_datetime_to_ft(ret, as_py_datetime)
+    if as_py_datetime is not None:
+        as_py_datetime = as_py_datetime.replace(tzinfo=GMT(0))
+        tr_datetime_to_ft(ret, as_py_datetime)
```
hoytak: What happens if as_py_datetime is None?  We should at least clear ret to make it clear what happens.

esamanas: I wasn't able to assign to ret from within that function. I couldn't figure out why. This is why I assigned ret to FLEX_UNDEFINED if a NaT was detected at line 1434, but I could just assign it regardless so that ret is always FLEX_UNDEFINED before this function is called. Then ret just won't be changed if as_py_datetime is None. Does that sound good?

PR #216, opened 2016-03-10
Allow cpp->unicode to be cpp->bytes if needed.

hoytak:
This is needed to get load_model to work in python 3.

### DIFF
--- a/oss_src/unity/python/sframe/cython/cy_cpp_utils.pyx
+++ b/oss_src/unity/python/sframe/cython/cy_cpp_utils.pyx
@@ -52,3 +52,28 @@ cdef string _attempt_cast_str_to_cpp(py_s) except *:
 
     # Okay, none of these worked, so error out.
     raise TypeError("Type '%s' cannot be interpreted as str." % str(type(py_s)))
+
+
+cdef bint _cpp_to_str_py3_decode_enabled = True
+
+cpdef disable_cpp_str_decode():
+    global _cpp_to_str_py3_decode_enabled
+    assert _cpp_to_str_py3_decode_enabled == True
+    _cpp_to_str_py3_decode_enabled = False
+
+cpdef enable_cpp_str_decode():
+    global _cpp_to_str_py3_decode_enabled
+    assert _cpp_to_str_py3_decode_enabled == False
+    _cpp_to_str_py3_decode_enabled = True
+
+cdef _cpp_to_str_py3_decode(const string& cpp_s):
+    """
+    Decodes a c++ string into bytes if disable_cpp_str_decode() has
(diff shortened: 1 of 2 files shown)


oss_src/unity/python/sframe/cython/cy_cpp_utils.pyx
```diff
@@ -52,3 +52,28 @@ cdef string _attempt_cast_str_to_cpp(py_s) except *:
...
+cdef bint _cpp_to_str_py3_decode_enabled = True
+
+cpdef disable_cpp_str_decode():
+    global _cpp_to_str_py3_decode_enabled
+    assert _cpp_to_str_py3_decode_enabled == True
+    _cpp_to_str_py3_decode_enabled = False
+
+cpdef enable_cpp_str_decode():
+    global _cpp_to_str_py3_decode_enabled
+    assert _cpp_to_str_py3_decode_enabled == False
```
ylow: drop the assert. otherwise this cannot be used recursively.

oss_src/unity/python/sframe/cython/cy_cpp_utils.pyx
```diff
@@ -52,3 +52,28 @@ cdef string _attempt_cast_str_to_cpp(py_s) except *:
...
+    _cpp_to_str_py3_decode_enabled = True
+
```
ylow: note that this is kind of iffy. Especially since the flag could be set, then a ctrl-C / some other failure happens which cause the flag to stay set. A "with xxx" mechanism is preferred.

hoytak: Yes, the guideline is to use this strictly in a try ... finally block.  But
let's do implement it as a with mechanism.  I'll revisit that.

<http://www.dato.com>   Hoyt Koepke
Data Scientist
hoytak@dato.com   <https://twitter.com/datoinc>

ylow:
@hoytak progress on this?

hoytak:
No; been deep in other stuff.  I'll try to get to it ASAP.  But it's not clear that this solves the problem fully until @TobyRoseman can confirm this patch works with all the python 2 and python 3 encoding issues throughout the codebase...

TobyRoseman:
@hoytak - not sure exactly what you want me to confirm. I can tell you this code has been checked into our private repository and no issues have been discovered.

hoytak:
Thanks @TobyRoseman.   That's exactly what I was hoping to hear. 

To make this permanent, we should set to be a with statement.  I'll try to do that tomorrow. 

-- Hoyt

ylow:
Closing this in preference of sync from internal repo.

PR #218, opened 2016-03-10
Invertible JSON serialization

znation:
This change adds an invertible (to/from) JSON serialization for SFrame,
SArray, and flexible_type that preserves the underlying type semantics
and values. This differs from the Python built-in json module
serialization in a few important ways:

* Handles SFrame and SArray (by converting to/from flexible_type
  invertibly).
* Handles all flexible_type (including datetime, array, Image)
* Does not necessarily handle any non-flexible_type value (boolean for
  instance).

The Python API is intended to match the Python built-in json module
(loads and dumps) but only handle "SFrame" types (SFrame, SArray, and
flexible_type).

The nitty gritty details:

* Adds a drop of RapidJSON under oss_src/rapidjson
* Adds fully invertible JSON serialization for flexible_type<->JSON in
  oss_src/unity/extensions/json.
* Adds Python wrappers to convert SFrame/SArray to and from
  flexible_type to allow them to be serialized using the above.
* Adds Python unit tests for both of the above.
* Fixes some minor bugs:
  * Non-invertible type names in flexible_type_base_types
  * Non-const read-only pointer in image_type

Change-Id: I3489af5ab4245b371797eea938ad49e7ed1ab9f3

### DIFF
--- a/oss_src/flexible_type/flexible_type_base_types.hpp
+++ b/oss_src/flexible_type/flexible_type_base_types.hpp
@@ -723,7 +723,7 @@ inline flex_type_enum flex_type_enum_from_name(const std::string& name) {
     {"float", flex_type_enum::FLOAT},
     {"string", flex_type_enum::STRING},
     {"array", flex_type_enum::VECTOR},
-    {"recursive", flex_type_enum::LIST},
+    {"list", flex_type_enum::LIST},
     {"dictionary", flex_type_enum::DICT},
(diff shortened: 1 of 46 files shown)


znation:
@ylow @haijieg @esamanas @srikris Please review, thanks!

Note to reviewers: you can mostly skip over the `rapidjson` code (this is just a drop of rapidjson headers into our repo). Please review the changes before/after (especially those that touch existing code).

PR #224, opened 2016-03-15
Detailed logging for pylambda_worker exceptions.

hoytak:
This is needed to track down lambda worker issues more intelligently.

### DIFF
--- a/oss_src/lambda/pylambda_worker.cpp
+++ b/oss_src/lambda/pylambda_worker.cpp
@@ -72,6 +72,8 @@ int EXPORT pylambda_worker_main(const std::string& root_path,
   LOG_DEBUG_WITH_PID("server_address = '" << server_address << "'");
   LOG_DEBUG_WITH_PID("parend pid = " << parent_pid);
 
+  size_t _last_line = __LINE__;
+#define __TRACK do { _last_line = __LINE__; } while(0)  
   try {
 
     LOG_DEBUG_WITH_PID("Library function entered successfully.");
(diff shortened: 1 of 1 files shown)


oss_src/lambda/pylambda_worker.cpp
```diff
@@ -72,6 +72,8 @@ int EXPORT pylambda_worker_main(const std::string& root_path,
   LOG_DEBUG_WITH_PID("server_address = '" << server_address << "'");
   LOG_DEBUG_WITH_PID("parend pid = " << parent_pid);
 
+  size_t _last_line = __LINE__;
+#define __TRACK do { _last_line = __LINE__; } while(0)
```
TobyRoseman: This is the C way of wrapping macros. From what I understand there is a more straight forward way to do this in C++. You can introduce arbitrary nesting where ever you like. So I believe you can just do:

`#define __TRACK do { _last_line = __LINE__; }`

hoytak: Yes, you can; however, it messes up the ending ";", which is half the
reason to do it the C way.  It preserves macros as terminated statements
with required ";".

<http://www.dato.com>   Hoyt Koepke
Data Scientist
hoytak@dato.com   <https://twitter.com/datoinc>

oss_src/lambda/pylambda_worker.cpp
```diff
@@ -81,49 +83,54 @@ int EXPORT pylambda_worker_main(const std::string& root_path,
...
-    return 104;
+    logstream(LOG_ERROR) << "PyLambda C++ Error: " << error.what()
+                         << "; last successful line =" << _last_line << std::endl;
+    return 2000 + _last_line;
   } catch (...) {
-    logstream(LOG_ERROR) << "Unknown PyLambda Error." << std::endl;
-    return 105;
+    logstream(LOG_ERROR) << "Unknown PyLambda Error"
+                         << "; last successful line =" << _last_line << std::endl;
+    return 3000 + _last_line;
```
ylow: Do any of these error codes here end up being returned as exit codes? Because exit codes are only in the range 0-255

hoytak: Ah, good catch.  Will change.

PR #229, opened 2016-03-16
Switched the sframe <-> JSON converter to use relative imports everywhere

ylow:


znation:
Looks good to me! Thanks for fixing that.

PR #230, opened 2016-03-16
Fix sarray_sorted_buffer size initialization

haijieg:
Also reduce the SGRAPH_VID_INGRESS_BUFFER from 3m to 1m.

This should fix the tree model.show() issue on windows with insufficient memory


hoytak:
This looks like a really obvious bugfix to me.

Issue #106, opened 2015-12-14
Docstring of SFrame.print_rows evaluates sys.stdout


rschindlauer:
The docstring signature of `SFrame.print_rows` is supposed to look like this:
```
def print_rows(self, num_rows=10, num_columns=40, max_column_width=30,
               max_row_width=80, output_file=sys.stdout):
```
However, looking at the docstring in the python shell it looks like `sys.stdout` somehow got evaluated:
```
print_rows(self, num_rows=10, num_columns=40, max_column_width=30,
           max_row_width=80, output_file=<open file '<stdout>', mode 'w'>) unbound sframe.data_structures.sframe.SFrame method
```

MatthewWest:
This may be unavoidable. Python evaluates default arguments once at definition time, rather than keeping them as pieces of code to be evaluated at call time (see [this](http://stackoverflow.com/questions/1651154/why-are-default-arguments-evaluated-at-definition-time-in-python) Stack Overflow question for more discussion).

One method which appears to be a solution would be monkey-patching the `__repr__` method on `sys.stdout`, but from a cursory examination it looks like this might not be possible (see [this](http://stackoverflow.com/questions/26050557/python-why-does-a-dynamically-added-repr-method-not-override-the-default) Stack Overflow question for details).

znation:
Fixed with 6216ad1

PR #237, opened 2016-03-24
Naive fix to #106

MatthewWest:
Rather than having sys.stdout get evaluated in the function definition,
use `None` as a default value and replace None with stdout in the
function body. Not sure this is the right solution...


hoytak:
Looks good to me.

hoytak:
Thanks for the contribution!

PR #268, opened 2016-04-19
A new version of spark_unity.jar that contains toDataFrame(...) funct…

esoroush:
A new version of spark_unity.jar that contains toDataFrame(...) functionality


TobyRoseman:
Looks good to me.

PR #273, opened 2016-04-25
Python 3.5 Support

TobyRoseman:
Biggest change is pushing a lot of our version of our dependencies. We hadn't done this in a while and most didn't support Python 3.5.

For backwards compatibility, I kept the `python3` switch building python 3.4. I'm happy to change that to `python3.4` if people thinks that better.


TobyRoseman:
This is now ready for review; @ylow please review.

ylow:
Surprisingly little changes to support py3.

PR #279, opened 2016-04-26
Improved .contains(), list to array.array casting. Operator code cleanup

ylow:
- sarray.contains() can now query inside a list, an array, and a dict (for keys)
 - Added an sarray.is_in() which is the inverse of sarray.contains()
 - We can now cast from list to array
 - Some code cleanup in operators. There was some really strange handling of vector types.


ylow:
@hershaw you might like this.

esamanas:
Looks good to me!

PR #284, opened 2016-04-27
Decouple JSON serializability from serialization

znation:
This change accomplishes a few goals in order to help preserve backward
compatibility with naive JSON serialization (passing something like
`list(sf)` directly into `json.dumps`) and consumers using such
serialization.

* Separates data and schema serialization (so that data can be sent
  schemaless if desired, with schema attached, or with schema sent
  separately, to support a backward-compatibility path for existing
  APIs).
* Separates serializability (conversion between rich types and a set
  of values that are naively serializable) from serialization
  (conversion between naively serializable and strings).
* Moves responsibility for serialization itself out of this API (and
  into something like Python's `json` module).
* Removes rapidjson dependency since serialization is now handled at the
  Python layer.
* Fixes a mismatch in flexible_type type name between NoneType and
  undefined.


oss_src/unity/extensions/json/decoder.cpp
```diff
@@ -0,0 +1,326 @@
...
+  for (const auto& kv : dict) {
+    if (kv.first == key) {
+      return kv.second;
+    }
+  }
+  std::stringstream msg;
+  msg << "Expected key \"";
+  msg << key;
+  msg << "\" was not present in dictionary input.";
+  throw msg.str();
```
ylow: Preferred is 
```
log_and_throw(...); 
```

it is included by
```
#include <logger/logger.hpp> 
```

oss_src/unity/extensions/json/decoder.cpp
```diff
@@ -0,0 +1,326 @@
...
+    case 0:
+      // flexible_type
+      return variant_get_value<flexible_type>(v);
+    default:
+      throw "Type mismatch: expected flexible_type as input to flex_from_variant.";
+  }
+}
+
+static flex_string _flex_string_from_variant(const variant_type& v) {
+  flexible_type f = _flexible_type_from_variant(v);
```
ylow: This function can probably be replaced with just a call to
```
variant_get_value<flex_string>(v)
```
It does raise an exception of something like "Expecting blah but we got a bleh" if it does not actually contain a string.

oss_src/unity/extensions/json/decoder.cpp
```diff
@@ -0,0 +1,326 @@
...
+    case flex_type_enum::DICT:
+      return _dict_from_serializable(data, schema);
+    case flex_type_enum::DATETIME:
+    case flex_type_enum::IMAGE:
+      throw "Unexpected input to _any_from_serializable: serializable flex_type does not include datetime or Image types.";
+  }
+}
+
+static schema_t schema_from_flex_dict(const flexible_type& f) {
+  flex_dict d = f.get<flex_dict>();
```
ylow: add an assert on d is flex_dict
```
ASSERT_EQUALS(f.get_type(), flex_type_enum::DICT)
```
its in
```
#include <logger/assertions.hpp>
```

ylow:
looks good. some minor changes.

znation:
Thanks for the review @ylow -- feedback addressed in 9941eb1ba391f22b3a76b16cb0da0ba9a6afe058

PR #286, opened 2016-04-29
Added a "sequence" operator.

ylow:
This will allow add_row_number and gl.SArray.from_sequence to be completely free.


haijieg:
how is this different than the range operator in operators/range.hpp?

ylow:
oh. indeed.... they are the same thing.

PR #287, opened 2016-04-29
Slice op

haijieg:
## Add slice opteration to sframe query planner
The slice operation can be directly applied to linear query plan,
or it will materialize (modifies) the existing query plan first.

Also add some utility function:
- is_linear_graph checks whether a query plan is linear
- make_sliced_graph slices the source node of a linear plan

## Make unity_sframe/unity_sarray::copy_range() more lazy
When step is 1, copy_range() will only sliced the query plan
instead of materializing. This affects all operations that
use copy_range(), e.g. and tail(), operator[], and make them much
faster.

- Fix the planner slice function to drop memoized length info.
- Change sframe callback implemenetation to use slice oeprator.

### DIFF
--- a/oss_src/sframe_query_engine/planning/planner.cpp
+++ b/oss_src/sframe_query_engine/planning/planner.cpp
@@ -51,7 +51,7 @@ static sframe execute_node_impl(pnode_ptr input_n, const materialize_options& ex
  * Executes a query plan, potentially parallelizing it if possible.
  * Also implements fast paths in the event the input node is a source node.
  */
-static sframe execute_node(pnode_ptr input_n, materialize_options exec_params) {
+static sframe execute_node(pnode_ptr input_n, const materialize_options& exec_params) {
   // fast path for SFRAME_SOURCE. If I am not streaming into
   // a callback, I can just call save
   if (exec_params.write_callback == nullptr &&
--- a/oss_src/sframe_query_engine/planning/planner.cpp
+++ b/oss_src/sframe_query_engine/planning/planner.cpp
@@ -31,7 +31,6 @@ REGISTER_GLOBAL(int64_t, SFRAME_MAX_LAZY_NODE_SIZE, true);
 static sframe execute_node_impl(pnode_ptr input_n, const materialize_options& exec_params) {
   // Either run directly, or split it up into a parallel section
   if(is_parallel_slicable(input_n) && (exec_params.num_segments != 0)) {
-
     size_t num_segments = exec_params.num_segments;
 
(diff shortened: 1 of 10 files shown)


oss_src/sframe_query_engine/planning/planner.cpp
```diff
@@ -52,7 +51,7 @@ static sframe execute_node_impl(pnode_ptr input_n, const materialize_options& ex
  * Executes a query plan, potentially parallelizing it if possible.
  * Also implements fast paths in the event the input node is a source node.
  */
-static sframe execute_node(pnode_ptr input_n, const materialize_options& exec_params) {
+static sframe execute_node(pnode_ptr input_n, materialize_options exec_params) {
```
ylow: why the change to copy?

oss_src/unity/extensions/additional_sframe_utilities.cpp
```diff
@@ -18,28 +19,59 @@ typedef int(*sframe_callback_type)(const graphlab::flexible_type*, size_t, void*
...
+  std::vector<std::vector<std::vector<flexible_type>>> rows_by_segment_id;
+  rows_by_segment_id.resize(nthreads);
+  auto copy_out_callback = [&](size_t segment_id,
+                               const std::shared_ptr<sframe_rows>& data) {
+    for (auto& row: (*data)) {
+      rows_by_segment_id[segment_id].push_back(row);
+    }
+    return false;
+  };
+  sliced_input.materialize_to_callback(copy_out_callback, nthreads);
```
ylow: this pulls the entire range into memory before triggering the callback. That might not be a good idea.

oss_src/unity/extensions/additional_sframe_utilities.cpp
```diff
@@ -18,28 +19,59 @@ typedef int(*sframe_callback_type)(const graphlab::flexible_type*, size_t, void*
...
```
ylow: Same as above

oss_src/sframe_query_engine/planning/planner.cpp
haijieg: good catch. It was modified when I was trying to change the execution component and I forget to change it back. Will fix.

oss_src/unity/extensions/additional_sframe_utilities.cpp
haijieg: Agreed. However, the benefit is that the materialization is now done in parallel with out requiring the callback function to be reentrent. Since the callback is ranged, the risk of out-of-memory is limited and controllable.

PR #293, opened 2016-05-10
External Columnar Sort

ylow:
The current sort algorithm we have has issues when the number of columns is very large since it performs a transpose of the data, and has problems with memory controls when there are columns with huge values. (for instance, images).

This proposes and implements a new sorting algorithm which makes use better of the columnar nature of the data. It switches back to the old sort implementation when there are few, small columns, but uses the new implementation for big columns. 

For instance, I tried using this to sort an SFrame of 1M rows, 1001 extremely sparse columns, and it completed in 2m 27s on my laptop without using additional storage; i.e. keeping within the 2GB cache limit on my laptop.  In the previous implementation this took over 10 mins, consumed 22GB of storage, then failed because I ran out of storage space.

The details of the new algorithm can be found in the header ec_sort.hpp


ylow:
Aargh. Failure. because my new sort returns the column order differently. To fix shortly.

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
```diff
@@ -0,0 +1,575 @@
...
+ *
+ * Returns an sframe. The last column of the sframe is the per-bucket 
+ * forward map.
+ */
+static sframe ec_scatter_partitions(sframe input,
+                                    size_t rows_per_bucket,
+                                    const std::vector<bool>& indirect_column,
+                                    std::shared_ptr<sarray<flexible_type> > forward_map) {
+  //  - For each (c,r,v) in data:
+  //    Write (c,v) to bucket `bucket of forward_map(r)`
```
TobyRoseman: Probably want to indent this line.

TobyRoseman:
This all seems reasonable to me. What does EC stand for?

ylow:
"External Columnar"

oss_src/sframe/sarray_file_format_v2.hpp
```diff
@@ -636,6 +636,48 @@ class sarray_group_format_writer_v2: public sarray_group_format_writer<T> {
...
+    DASSERT_LT(columnid, m_column_buffers.size());
+    DASSERT_EQ(m_array_open, true);
+
+    auto el_before_flush = m_column_buffers[columnid].elements_before_flush;
+    auto& buffer = m_column_buffers[columnid].segment_data[segmentid];
+    for(const auto& elem: t) {
+      buffer.push_back(elem);
+      if (buffer.size() >= el_before_flush) {
+        flush_block(columnid, segmentid);
+        el_before_flush = m_column_buffers[columnid].elements_before_flush;
```
haijieg: m_column_buffers[columnid].elements_before_flush changed by flush_block()?

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
```diff
@@ -0,0 +1,583 @@
...
+ * Scatters the input into a collection of buckets using the forward_map.
+ * The forward_map must be an SArray of the same length as the input and contain
+ * every integer from 0 to len-1.
+ *
+ * Returns an sframe. The last column of the sframe is the per-bucket 
+ * forward map.
+ */
+static sframe ec_scatter_partitions(sframe input,
+                                    size_t rows_per_bucket,
+                                    const std::vector<bool>& indirect_column,
```
haijieg: what's indirect_column?

oss_src/sframe_query_engine/algorithm/ec_sort.cpp
```diff
@@ -0,0 +1,195 @@
...
+  std::set<size_t> value_column_indices_set(value_column_indices.begin(), 
+                                            value_column_indices.end());
+  std::vector<flex_type_enum> value_column_types = infer_planner_node_type(value_columns);
+  {
+    std::set<flex_type_enum> value_column_type_set(value_column_types.begin(), 
+                                                   value_column_types.end());
+    // remove all the definitely small value types
+    value_column_type_set.erase(flex_type_enum::INTEGER);
+    value_column_type_set.erase(flex_type_enum::FLOAT);
+    value_column_type_set.erase(flex_type_enum::DATETIME);
```
haijieg: is value_column_type_set used anywhere?

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
```diff
@@ -0,0 +1,583 @@
...
+    logstream(LOG_INFO) << "Processing rows " 
+                        << forward_map_start 
+                        << " to " << forward_map_end << std::endl;
+    forward_map_reader->read_rows(forward_map_start, forward_map_end, 
+                                  forward_map_buffer);
+
+    // now in parallel over columns.
+    atomic<size_t> col_number = 0;
+    in_parallel(
+        [&](size_t unused, size_t unused2) {
```
haijieg: what are unused and unused2

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+            if (indirect_column[column_id]) {
+              // if this is to be an indirect column, 
+              // we use write the row number as the value
+              for (size_t actual_row = forward_map_start; actual_row < forward_map_end; ++actual_row) {
+                size_t output_row = 
+                    forward_map_buffer[actual_row - forward_map_start].get<flex_int>();
+                size_t output_segment = output_row / rows_per_bucket;
+                if (output_segment >= num_buckets) output_segment = num_buckets - 1;
+                writer->write_segment(column_id, output_segment, flexible_type(actual_row));
```
haijieg: is this the per-bucket forward map? The documentation says only the last column is the per-bucket forward map

ylow: The forward_map_buffer is read a few lines before for the whole bucket.

oss_src/sframe_query_engine/algorithm/ec_sort.cpp
ylow: Oops. That's an error. A few lines below should be value_column_type_set.size()

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
ylow: This is the optimiation as stated in ec_sort.hpp:
```
- Particular excessively heavy columns (like super big dictionaries,
+lists, images) , we store the row number instead during the scatter
+phase. And we use seeks against the original input to transfer it
+straight from input to output. We could take a guess about the
+column data size and use the seek strategy if it is too large.
```

oss_src/sframe/sarray_file_format_v2.hpp
ylow: Yes. Though, I could just change el_before_flush to be a reference and I don't need this line.

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
haijieg: Separate question: if there are many indirect columns, do we need to write the same row number that many times?

ylow: Technically no. Its a possible optimization. Though probably not too significant an optimization since the row numbers are small and the indirect columns are  big.

oss_src/sframe_query_engine/algorithm/ec_permute.cpp
```diff
@@ -0,0 +1,583 @@
...
+  //  In the implementation, each bucket corresponds to one segment.
+  
+  logstream(LOG_INFO) << "input size " << input.size() << std::endl;
+  logstream(LOG_INFO) << "forward map size " << forward_map->size() << std::endl;
+  input = input.add_column(forward_map);
+  auto num_buckets = (input.size() + rows_per_bucket - 1)/ rows_per_bucket;
+  sframe output;
+  auto out_column_types = input.column_types();
+  for (size_t i = 0;i < out_column_types.size(); ++i) {
+    if (indirect_column[i]) out_column_types[i] = flex_type_enum::INTEGER;
```
haijieg: what's the size of indirect_column? I'm asking because the output has one more column than the input (when its passed in). Worth putting a comment here.

haijieg:
LGTM. Please fix the issues found during the review. Nice work!!

PR #295, opened 2016-05-10
Fixes resource leak when saving sframes

ylow:
Remember to close columns after opening them.


ylow:
Invalid. There was not a leak.

PR #296, opened 2016-05-10
Fast blockwise parallel sarray iterator

hoytak:
This PR adds a new sarray iterator that runs through data block by
block for fast sequential access.


hoytak:
Will be pulled in through a massive merge of closed-source development into SFrame.

PR #300, opened 2016-05-11
Allowed from_const to take a type.

ylow:
This is very useful for creating an SArray of Nones.

### DIFF
--- a/oss_src/unity/python/sframe/test/test_sarray.py
+++ b/oss_src/unity/python/sframe/test/test_sarray.py
@@ -1537,6 +1537,18 @@ class SArrayTest(unittest.TestCase):
         self.assertEquals(list(g), [0] * 100)
         self.assertEqual(g.dtype(), int)
 
+        g = SArray.from_const(None, 100, float)
+        self.assertEquals(list(g), [None] * 100)
+        self.assertEqual(g.dtype(), float)
+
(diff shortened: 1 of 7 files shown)


oss_src/unity/python/sframe/test/test_sarray.py
```diff
@@ -1525,6 +1525,18 @@ def test_from_const(self):
...
+        g = SArray.from_const(None, 100, str)
+        self.assertEquals(list(g), [None] * 100)
+        self.assertEqual(g.dtype(), str)
+
+        g = SArray.from_const(0, 100, float)
+        self.assertEquals(list(g), [0.0] * 100)
+        self.assertEqual(g.dtype(), float)
+
+        g = SArray.from_const(0.0, 100, int)
+        self.assertEquals(list(g), [0] * 100)
```
srikris: Can you test the following

```
g = SArray.from_const(None, 100, float)
self.assertEquals(list(g), [None] * 100)
self.assertEqual(g.dtype(), float)

g = SArray.from_const(None, 100, int)
self.assertEquals(list(g), [None] * 100)
self.assertEqual(g.dtype(), int)

g = SArray.from_const(None, 100, list)
self.assertEquals(list(g), [None] * 100)
self.assertEqual(g.dtype(), list)
```

PR #301, opened 2016-05-12
Loading environment variables from a config file

hoytak:
This PR allows the user to permanently write environment variables to
a common config file, then have those values be persistent across
sessions.

Use sys_util.get_config_file() to return the name of this config file.
Use sys_util.write_config_file_value(k, v) to write a value to this config file.

### DIFF
--- a/oss_src/unity/python/sframe/sys_util.py
+++ b/oss_src/unity/python/sframe/sys_util.py
@@ -495,22 +495,25 @@ def setup_environment_from_config_file():
 
     if not exists(config_file):
         return
-    
-    config = _ConfigParser.SafeConfigParser()
-    config.read(config_file)
 
-    __section = "Environment"
-    
-    if config.has_section(__section):
-        items = config.items(__section)
+    try:
+        config = _ConfigParser.SafeConfigParser()
+        config.read(config_file)
 
-        for k, v in items:
-            try:
-                os.environ[k.upper()] = v
-            except Exception as e:
-                print(("WARNING: Error setting environment variable "
-                       "'%s = %s' from config file '%s': %s.")
-                      % (k, str(v), config_file, str(e)) )
+        __section = "Environment"
+
+        if config.has_section(__section):
+            items = config.items(__section)
+
+            for k, v in items:
+                try:
+                    os.environ[k.upper()] = v
+                except Exception as e:
+                    print(("WARNING: Error setting environment variable "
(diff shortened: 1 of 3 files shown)


oss_src/unity/python/sframe/sys_util.py
```diff
@@ -439,3 +442,98 @@ def _get_expanded_classpath(classpath):
                     for path in classpath.split(os.path.pathsep))
     logging.getLogger(__name__).debug('classpath being used: %s' % jars)
     return jars
+
+def get_library_name():
+    """
+    Returns either sframe or graphlab depending on which library
+    this file is bundled with.
+    """
+    from os.path import split, abspath, expanduser, join
```
TobyRoseman: not using `expanduser` or `join`

oss_src/unity/python/sframe/sys_util.py
```diff
@@ -439,3 +442,98 @@ def _get_expanded_classpath(classpath):
...
+    return __lib_name
+
+
+def get_config_file():
+    """
+    Returns the file name of the config file from which the environment
+    variables are written.
+    """
+    import os
+    from os.path import split, abspath, expanduser, join, exists
```
TobyRoseman: not using `split`

oss_src/unity/python/sframe/test/test_environment_config.py
```diff
@@ -0,0 +1,82 @@
...
+    sys.exit(0)
+else:
+    print("boo: GRAPHLAB_FILEIO_MAXIMUM_CACHE_CAPACITY = ", var, "and this is wrong.  Seriously.")
+    sys.exit(1)
+    
+""" % {"library" : get_library_name()}
+
+        run_file = join(test_dir, "run_test.py")
+
+        open(run_file, 'w').write(run_script)
```
TobyRoseman: Isn't there a risk the file might not be closed/flushed before it's executed?

oss_src/unity/python/sframe/sys_util.py
```diff
@@ -439,3 +442,98 @@ def _get_expanded_classpath(classpath):
...
+    """
+
+    from os.path import exists
+    
+    config_file = get_config_file()
+
+    if not exists(config_file):
+        return
+    
+    config = _ConfigParser.SafeConfigParser()
```
ylow: Do a big try except here. Just in case issues like
 - file exists but no read permissions
 - malformed file.

PR #304, opened 2016-05-12
Ternary operator

ylow:
Implements a ternary operator. A little like np.where. Extremely useful for replacing values in an SArray.

```
Returns an SArray with the same values as g with values above 10
clipped to 10

>>> g = SArray([6,7,8,9,10,11,12,13])
>>> SArray.where(g > 10, 10, g)
dtype: int
Rows: 8
[6, 7, 8, 9, 10, 10, 10, 10]

Returns an SArray with the same values as g with values below 10
clipped to 10

>>> SArray.where(g > 10, g, 10)
dtype: int
Rows: 8
[10, 10, 10, 10, 10, 11, 12, 13]

Returns an SArray with the same values of g with all values == 1
replaced by None

>>> g = SArray([1,2,3,4,1,2,3,4])
>>> SArray.where(g == 1, None, g)
dtype: int
Rows: 8
[None, 2, 3, 4, None, 2, 3, 4]
```

Also fixes a bunch of oddities with SArray.from_const.
Previously,
```
gl.SArray.from_const([1], 10, list)
```
will produce an SArray of
[[1.0],[1.0],[1.0],[1.0],[1.0]...]
converting each of the integers to a float.

Now it does the right thing.

### DIFF
--- a/oss_src/sframe_query_engine/operators/ternary_operator.hpp
+++ b/oss_src/sframe_query_engine/operators/ternary_operator.hpp
@@ -31,7 +31,7 @@ class operator_impl<planner_node_type::TERNARY_OPERATOR> : public query_operator
   
   planner_node_type type() const { return planner_node_type::TERNARY_OPERATOR; } 
 
-  static std::string name() { return "binary_transform"; }
+  static std::string name() { return "ternary"; }
         
   static query_operator_attributes attributes() {
     query_operator_attributes ret;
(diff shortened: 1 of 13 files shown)


ylow:
@hershaw the ternary operator we discussed.

hershaw:
This is great! Simpler and more flexible than the numpy implementation.

oss_src/sframe_query_engine/operators/ternary_operator.hpp
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  static std::shared_ptr<planner_node> make_planner_node(
+      std::shared_ptr<planner_node> condition,
+      std::shared_ptr<planner_node> istrue,
+      std::shared_ptr<planner_node> isfalse) {
+    
+    return planner_node::make_shared(planner_node_type::TERNARY_OPERATOR, 
+                                     {},
```
haijieg: what's the type of this {}? clang might have warning/error for std::map

ylow: didn't seem to on my mac. But I can make it explicit.

ylow:
@haijieg updated.

haijieg:
LGTM

PR #306, opened 2016-05-12
Add early termination support for .any() and .all()

### LINKED ISSUE #235: Early stopping for .any() and .all()
Some operators should stop early, but they do not currently due to the coroutines method of the lazy evaluation framework.   For example, see https://github.com/dato-code/SFrame/blob/master/oss_src/unity/lib/unity_sarray.cpp#L721.

ylow:
Fixes #235

### DIFF
--- a/oss_src/unity/lib/unity_sarray.cpp
+++ b/oss_src/unity/lib/unity_sarray.cpp
@@ -705,30 +705,36 @@ size_t unity_sarray::num_missing() {
 
 bool unity_sarray::all() {
   log_func_entry();
-  
+  class early_termination{};
   auto reductionfn = [](const flexible_type& f, int& segment_all)->void {
     segment_all &= !f.is_zero();
+    if (segment_all == 0) {
+      throw early_termination();
+    }
   };
-
-  return query_eval::reduce<int>(m_planner_node, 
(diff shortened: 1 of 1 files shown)


oss_src/unity/lib/unity_sarray.cpp
```diff
@@ -705,30 +705,36 @@ size_t unity_sarray::num_missing() {
 
 bool unity_sarray::all() {
   log_func_entry();
-  
+  class early_termination{};
```
hoytak: To make clearer what's going on, maybe this should be early_termination_return_zero?

oss_src/unity/lib/unity_sarray.cpp
```diff
@@ -705,30 +705,36 @@ size_t unity_sarray::num_missing() {
...
+                                   reductionfn, reductionfn, 1) > 0;
+  } catch (early_termination) {
+    return false;
+  }
 }
 
 bool unity_sarray::any() {
   log_func_entry();
-  auto reductionfn = [](const flexible_type& f, int& segment_all)->bool {
+  class early_termination{};
```
hoytak: ditto

PR #307, opened 2016-05-12
Add SArray.hash() and SArray.random_integers()

ylow:
These are very useful for things like exact subsampling, sorting, etc.


ylow:
URGH. git oddities. going to re-create this PR

PR #308, opened 2016-05-12
Add SArray.hash() and SArray.random_integers()

ylow:
These are very useful for things like exact subsampling, sorting, etc.

### DIFF
--- a/oss_src/unity/lib/unity_sarray.cpp
+++ b/oss_src/unity/lib/unity_sarray.cpp
@@ -1628,6 +1628,16 @@ std::shared_ptr<unity_sarray_base> unity_sarray::sample(float percent,
   return logical_filter(seq);
 }
 
+
+std::shared_ptr<unity_sarray_base> unity_sarray::hash(int random_seed) {
+  flex_int seed_hash = flexible_type((flex_int)(random_seed)).hash();
+  auto filter_fn = [seed_hash](const flexible_type& val)->flexible_type {
+        return hash64(val.hash() ^ seed_hash);
+      };
+  return transform_lambda(filter_fn, flex_type_enum::INTEGER, false, 0);
+}
+
+
 std::shared_ptr<unity_sarray_base>
 unity_sarray::count_bag_of_words(std::map<std::string, flexible_type> options) {
   log_func_entry();
--- a/oss_src/unity/lib/unity_sarray.hpp
+++ b/oss_src/unity/lib/unity_sarray.hpp
@@ -515,6 +515,12 @@ class unity_sarray: public unity_sarray_base {
    */
   std::shared_ptr<unity_sarray_base> sample(float percent, int random_seed);
 
+  /**
+   * Returns an SArray of type flex_int that contains the hash of each element.
+   * The hash function takes a seed value so this can be used for
+   * random generation as well.
+   */
+  std::shared_ptr<unity_sarray_base> hash(int seed);
 
   /**
    * Do a word-count for each element in the SArray and return a SArray of dictionary
--- a/oss_src/unity/python/sframe/cython/cy_sarray.pxd
+++ b/oss_src/unity/python/sframe/cython/cy_sarray.pxd
@@ -64,6 +64,7 @@ cdef extern from "<unity/lib/api/unity_sarray_interface.hpp>" namespace 'graphla
         unity_sarray_base_ptr drop_missing_values() except +
         unity_sarray_base_ptr fill_missing_values(flexible_type) except +
         unity_sarray_base_ptr sample(float, int) except +
+        unity_sarray_base_ptr hash(int) except +
         void materialize() except +
         bint is_materialized() except +
         unity_sarray_base_ptr append(unity_sarray_base_ptr) except +
@@ -176,6 +177,8 @@ cdef class UnitySArrayProxy:
     cpdef fill_missing_values(self, default_value)
 
     cpdef sample(self, float percent, int seed)
+
+    cpdef hash(self, int seed)
 
     cpdef materialize(self)
 
--- a/oss_src/unity/python/sframe/data_structures/sarray.py
+++ b/oss_src/unity/python/sframe/data_structures/sarray.py
@@ -1980,13 +1980,13 @@ class SArray(object):
             return SArray(_proxy=self.__proxy__.hash(seed))
 
     @classmethod
-    def random_integers(self, size, seed=None):
+    def random_integers(cls, size, seed=None):
         """
         Returns an SArray with random integer values.
         """
         if seed is None:
             seed = abs(hash("%0.20f" % time.time())) % (2 ** 31)
-        return SArray.from_sequence(size).hash(seed)
+        return cls.from_sequence(size).hash(seed)
 
     def _save_as_text(self, url):
         """
--- a/oss_src/unity/python/sframe/data_structures/sarray.py
+++ b/oss_src/unity/python/sframe/data_structures/sarray.py
@@ -1855,8 +1855,8 @@ class SArray(object):
         dryrun = [fn(i) for i in self.head(100) if i is not None]
         if dtype == None:
             dtype = infer_type_of_list(dryrun)
-        if not seed:
-            seed = time.time()
+        if seed is None:
+            seed = abs(hash("%0.20f" % time.time())) % (2 ** 31)
 
         # log metric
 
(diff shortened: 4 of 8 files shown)


oss_src/unity/lib/unity_sarray.cpp
```diff
@@ -1622,6 +1622,16 @@ std::shared_ptr<unity_sarray_base> unity_sarray::sample(float percent,
   return logical_filter(seq);
 }
 
+
+std::shared_ptr<unity_sarray_base> unity_sarray::hash(int random_seed) {
+  flex_int seed_hash = flexible_type((flex_int)(random_seed)).hash();
+  auto filter_fn = [seed_hash](const flexible_type& val)->flexible_type {
+        return hash64(val.hash() ^ seed_hash);
```
hoytak: Do hash64(val.hash(), seed_hash); it's set up for this.

oss_src/unity/lib/unity_sarray.cpp
```diff
@@ -1622,6 +1622,16 @@ std::shared_ptr<unity_sarray_base> unity_sarray::sample(float percent,
   return logical_filter(seq);
 }
 
+
+std::shared_ptr<unity_sarray_base> unity_sarray::hash(int random_seed) {
+  flex_int seed_hash = flexible_type((flex_int)(random_seed)).hash();
```
hoytak: just do hash64(random_seed)

oss_src/unity/lib/unity_sarray.cpp
```diff
@@ -1622,6 +1622,16 @@ std::shared_ptr<unity_sarray_base> unity_sarray::sample(float percent,
   return logical_filter(seq);
 }
 
+
+std::shared_ptr<unity_sarray_base> unity_sarray::hash(int random_seed) {
```
hoytak: why int?  Shouldn't this be size_t?

oss_src/unity/lib/unity_sarray.hpp
```diff
@@ -515,6 +515,12 @@ class unity_sarray: public unity_sarray_base {
    */
   std::shared_ptr<unity_sarray_base> sample(float percent, int random_seed);
 
+  /**
+   * Returns an SArray of type flex_int that contains the hash of each element.
+   * The hash function takes a seed value so this can be used for
+   * random generation as well.
+   */
+  std::shared_ptr<unity_sarray_base> hash(int seed);
```
hoytak: ditto - int vs. size_t.

oss_src/unity/python/sframe/cython/cy_sarray.pxd
```diff
@@ -64,6 +64,7 @@ cdef extern from "<unity/lib/api/unity_sarray_interface.hpp>" namespace 'graphla
         unity_sarray_base_ptr drop_missing_values() except +
         unity_sarray_base_ptr fill_missing_values(flexible_type) except +
         unity_sarray_base_ptr sample(float, int) except +
+        unity_sarray_base_ptr hash(int) except +
```
hoytak: ditto

oss_src/unity/python/sframe/cython/cy_sarray.pxd
```diff
@@ -175,6 +176,8 @@ cdef class UnitySArrayProxy:
 
     cpdef sample(self, float percent, int seed)
 
+    cpdef hash(self, int seed)
```
hoytak: ditto. or just do a python object, then call hash() on that.  then it's a 64bit int.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -1873,6 +1873,36 @@ def sample(self, fraction, seed=None):
...
+        with cython_context():
+            return SArray(_proxy=self.__proxy__.hash(seed))
+
+    @classmethod
+    def random_integers(self, size, seed=None):
+        """
+        Returns an SArray with random integer values.
+        """
+        if seed is None:
+            seed = time.time()
```
hoytak: This has resolution up to a second, so calls within the same second produce the same values.  Use ` seed = hash("%0.20f" % time.time())`.

oss_src/unity/python/sframe/test/test_sarray.py
```diff
@@ -977,6 +977,31 @@ def test_sample(self):
         sa_sample = SArray().sample(.5, 9)
         self.assertEqual(len(sa_sample), 0)
 
+    def test_hash(self):
```
hoytak: Add test that two subsequent calls with seed = None do not have the same values.

ylow: hash is fully deterministic. seed=None does not make sense for a hash method.

oss_src/unity/lib/unity_sarray.cpp
ylow: Was going to. But we have "int" *everywhere* going quite deep. sample takes an int as well.
Fixing that will be a much larger code change than I want to do now.

oss_src/unity/lib/unity_sarray.hpp
ylow: Can't change to size_t now. int goes quite deep.

oss_src/unity/lib/unity_sarray.cpp
hoytak: Ah, okay.  Let's maybe then take it and do `seed = hash(seed) % 2**31` at the python level.

oss_src/unity/python/sframe/data_structures/sarray.py
ylow: At the moment, this is consistent with how sa.sample() works.I could fix it there. But then it will definitely be necessary to expand int to size_t. (or ssize_t). which will be a large change.

oss_src/unity/lib/unity_sarray.cpp
hoytak: Reference: https://github.com/dato-code/SFrame/blob/master/oss_src/unity/python/sframe/data_structures/sframe.py#L3241

oss_src/unity/python/sframe/data_structures/sarray.py
hoytak: Ref: https://github.com/dato-code/SFrame/blob/master/oss_src/unity/python/sframe/data_structures/sframe.py#L3241

oss_src/unity/python/sframe/test/test_sarray.py
hoytak: Ah, sorry, I meant on random_integers.

oss_src/unity/lib/unity_sarray.cpp
ylow: Oddly: https://github.com/dato-code/SFrame/blob/master/oss_src/unity/python/sframe/data_structures/sframe.py#L3192

ylow: Then sample should probably be fixed too.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -1952,12 +1952,42 @@ def sample(self, fraction, seed=None):
...
+        -------
+        out : SArray
+            An integer SArray with a hash value for each element. Identical
+            elements are hashed to the same value
+        """
+        with cython_context():
+            return SArray(_proxy=self.__proxy__.hash(seed))
+
+    @classmethod
+    def random_integers(self, size, seed=None):
```
TobyRoseman: When it's a classmethod, the first parameter shouldn't be `self`, usually `cls` is used.

TobyRoseman: Maybe `length` rather than `size`.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -1952,12 +1952,42 @@ def sample(self, fraction, seed=None):
...
+        """
+        Returns an SArray with random integer values.
+        """
+        if seed is None:
+            seed = abs(hash("%0.20f" % time.time())) % (2 ** 31)
+        return SArray.from_sequence(size).hash(seed)
```
TobyRoseman: Assuming you take my previous comment you probably want to change "SArray" to "cls". At any rate I don't think you should have "SArray" hard coded.

ylow:
@TobyRoseman Yup. done

TobyRoseman:
Looks good to me

PR #310, opened 2016-05-12
Robustification of lambda worker startup code.

hoytak:
Makes the starting code to the lambda worker process more robust. 

env variable GRAPHLAB_LAMBDA_WORKER_CONNECTION_TIMEOUT now controls the timeout. Default is 60 seconds.  Setting to 0 means there is one try.  Setting to -1 means no timeout.

### DIFF
--- a/oss_src/lambda/worker_pool.hpp
+++ b/oss_src/lambda/worker_pool.hpp
@@ -17,8 +17,12 @@
 #include<parallel/pthread_tools.hpp>
 #include<process/process.hpp>
 #include<cppipc/client/comm_client.hpp>
+#include<timer/timer.hpp>
 
 namespace graphlab {
+
+extern double LAMBDA_WORKER_CONNECTION_TIMEOUT;
+
 namespace lambda {
 
 /**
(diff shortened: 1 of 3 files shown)


oss_src/lambda/worker_pool.hpp
```diff
@@ -91,43 +95,114 @@ std::unique_ptr<worker_process<ProxyType>> spawn_worker(std::vector<std::string>
...
-    // Make sure the server process is still around
+
+  auto check_process_exists = [&]() -> bool {
+    // Make sure the server process is still around.
     if (!new_process->exists()) {
-      logstream(LOG_ERROR) << "Lambda worker process terminated unexpectedly" << std::endl;
-      break;
+      int ret_code = new_process->get_return_code();
+
+      std::ostringstream ss;
```
ylow: Why not just stream into logstream(LOG_ERROR) directly? The stringstream is not needed.

oss_src/lambda/worker_pool.hpp
```diff
@@ -91,43 +95,114 @@ std::unique_ptr<worker_process<ProxyType>> spawn_worker(std::vector<std::string>
...
+    } else {
+      return true;
     }
+  };
+
+  auto log_exception = [&](const std::string& e) {
+    std::ostringstream ss;
+    ss << "Error starting CPPIPC connection in connecting to lambda worker at "
+       << worker_address
+       << " (conn_time = " << conn_timer.current_time()
```
ylow: ditto

oss_src/lambda/worker_pool.hpp
```diff
@@ -91,43 +95,114 @@ std::unique_ptr<worker_process<ProxyType>> spawn_worker(std::vector<std::string>
...
+       << "; attempt count = " << retry << "): " << e;
+    logstream(LOG_ERROR) << ss.str() << std::endl;
+  };
+
+  // The actual timeout is controlled by conn_timer, so that should be
+  // what causes the loop to exit in case of failure.  However, put in
+  // a really large number here in case of infinite-loop programming
+  // errors.
+  while (retry < 1000000) {
```
ylow: eh... just do while(1)

oss_src/lambda/worker_pool.hpp
```diff
@@ -91,43 +95,114 @@ std::unique_ptr<worker_process<ProxyType>> spawn_worker(std::vector<std::string>
...
+      std::unique_ptr<cppipc::comm_client> tmp_cli(
+          new cppipc::comm_client({}, worker_address, connection_timeout));
+
       cppipc::reply_status status = tmp_cli->start();
+
       if (status == cppipc::reply_status::OK) {
         // Success
-        logstream(LOG_INFO) << "Connected to worker at " << worker_address << std::endl;
+        std::ostringstream ss;
+        ss  << "Connected to worker " << new_process->get_pid()
```
ylow: stringstream not needed.

oss_src/lambda/worker_pool.hpp
hoytak: It isn't locking in a multithreaded environment, so you get interleaving of the output if you have a lot of << parts.   A single << works fine though.

PR #311, opened 2016-05-13
Added use of set_new_handler

ylow:
See http://en.cppreference.com/w/cpp/memory/new/set_new_handler

If a memory allocation failure occurs, we get a chance to flush caches to
free up enough memory so that the allocation may succeed after that.


hoytak:
I'd say let's get it in and see what happens.  Not sure how to test this either.

PR #317, opened 2016-05-19
Bugfix for slicing optimization operator error.

hoytak:


oss_src/sframe_query_engine/planning/optimizations/union_transforms.hpp
```diff
@@ -99,10 +104,23 @@ class opt_union_on_source : public opt_union_transform {
...
           size_t end_index   = n->inputs[i]->p("end_index");
 
-          distinct_input_ranges.insert({begin_index, end_index});
+          size_t size = 0;
+
+          if(t == planner_node_type::SFRAME_SOURCE_NODE) {
+            // Get the size of the sframe
+            size = n->inputs[i]->any_p<sframe>("sframe").num_rows();
+          } else {
+            // Get the size of the sarray
```
znation: Is it possible `planner_node_type` will ever have values other than sframe and sarray? When traversing enums I like to use:

```
switch (enum_value) {
    case value_1:
        // logic here
        break;
    case value_2:
        // logic here
        break;
}
```

If you omit a default path, you get a compiler warning/error if the enum ever changes to have more values since you'll no longer cover all the cases.

znation:
Not sure I understand the unit test code, but the change generally looks good to me.

oss_src/sframe_query_engine/planning/optimizations/union_transforms.hpp
hoytak: Yes, it can have dozens of types, but the others are excluded (they don't apply to this transform) by the if statement in lines 96/101 and 97/102.  Both of these have a common set of parameters, minus retrieving the size.

znation: Ah, thanks! I didn't see those lines (GitHub collapsed them away).

ylow:
Can you add that minimal python repro as a test case?

oss_src/unity/python/sframe/test/test_sframe.py
```diff
@@ -3306,6 +3306,12 @@ def test_materialize(self):
         sf.materialize()
         self.assertTrue(sf.is_materialized())
 
+    def test_materialization_slicing(self):
+        # Has been known to fail.
+        g=SFrame({'a':range(100)})[:10]
+        g['b'] = g['a'] + 1
+        g['b'].materialize()
+
```
ylow: failure happens on a g.materialize() after this

ylow:
ah you did. but you need one more line in the test case.

PR #318, opened 2016-05-20
Fix memory leak in png_encode

haijieg:
The image_type uses boost::shared_ptr<char[]> to manage the image bytearray
in which the pointer is required to be allocated using C++ new char[].

In the png encoder, we use malloc/realloc. This fixes the leak by
makeing a copy of the png encoded memory buffer into a "new" allocated
array.


ylow:
+1

PR #319, opened 2016-05-21
fixes a segfault at termination of optimizations.cxxtest

ylow:
The global variable sliced_graph_memo causes problems because it holds on to
handles which are released after the global cache manager is released.


ylow:
Also @znation

znation:
Looks good to me!

PR #324, opened 2016-05-24
Add SArray.shape property

znation:
This mirrors the SFrame.shape property and gives API compatibility with
Pandas series.


hoytak:
Looks good, but can we put a test in test_sarray.py ?

znation:
@hoytak Sure, I will do that. Thanks!

PR #326, opened 2016-05-24
Add SArray.random_split

znation:
This behaves the same as SFrame.random_split, but on a single column.

### DIFF
--- a/oss_src/unity/python/sframe/data_structures/sarray.py
+++ b/oss_src/unity/python/sframe/data_structures/sarray.py
@@ -2900,6 +2900,44 @@ class SArray(object):
         with cython_context():
             return SArray(_proxy = self.__proxy__.item_length())
 
+    def random_split(self, fraction, seed=None):
+        """
+        Randomly split the rows of an SArray into two SArrays. The first SArray
+        contains *M* rows, sampled uniformly (without replacement) from the
+        original SArray. *M* is approximately the fraction times the original
+        number of rows. The second SArray contains the remaining rows of the
+        original SArray.
+
+        Parameters
+        ----------
+        fraction : float
+            Approximate fraction of the rows to fetch for the first returned
+            SArray. Must be between 0 and 1.
+
+        seed : int, optional
+            Seed for the random number generator used to split.
+
+        Returns
+        -------
+        out : tuple [SArray]
+            Two new SArrays.
+
+        Examples
+        --------
+        Suppose we have an SArray with 1,024 rows and we want to randomly split
+        it into training and testing datasets with about a 90%/10% split.
+
+        >>> sa = graphlab.SArray(range(1024))
+        >>> sa_train, sa_test = sa.random_split(.9, seed=5)
+        >>> print(len(sa_train), len(sa_test))
+        922 102
+        """
+        from .sframe import SFrame
+        temporary_sf = SFrame()
+        temporary_sf['X1'] = self
+        (train, test) = temporary_sf.random_split(fraction, seed)
+        return (train['X1'], test['X1'])
+
     def split_datetime(self, column_name_prefix = "X", limit=None, tzone=False):
         """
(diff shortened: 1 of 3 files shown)


oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -2885,6 +2885,44 @@ def item_length(self):
...
+        -------
+        out : tuple [SArray]
+            Two new SArrays.
+
+        Examples
+        --------
+        Suppose we have an SArray with 1,024 rows and we want to randomly split
+        it into training and testing datasets with about a 90%/10% split.
+
+        >>> sf = graphlab.SFrame({'id': range(1024)})
```
TobyRoseman: The example should be using `SArray` not `SFrame`, right?

znation: Good catch. Thanks, I will fix.

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -2885,6 +2885,44 @@ def item_length(self):
...
+        >>> sf_train, sf_test = sf.random_split(.9, seed=5)
+        >>> print len(sf_train), len(sf_test)
```
TobyRoseman: Python 3 compatible code is best: `print(len(sf_train), len(sf_test))`

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -2885,6 +2885,44 @@ def item_length(self):
...
+        922 102
+        """
+        import sframe
```
TobyRoseman: This will fail when we pull this code back into GLC. The package name is no longer `sframe`. We have a helper method for returning the package, not sure what it's called I can look it up if you want.

oss_src/unity/python/sframe/data_structures/sarray.py
znation: Thanks, I will update both here and for the SFrame docstring.

oss_src/unity/python/sframe/data_structures/sarray.py
znation: Oh, right! Thanks, I will fix that as well.

hoytak:
Looks good to me!

oss_src/unity/python/sframe/data_structures/sarray.py
```diff
@@ -2885,6 +2885,44 @@ def item_length(self):
...
+        --------
+        Suppose we have an SArray with 1,024 rows and we want to randomly split
+        it into training and testing datasets with about a 90%/10% split.
+
+        >>> sa = graphlab.SArray(range(1024))
+        >>> sa_train, sa_test = sa.random_split(.9, seed=5)
+        >>> print(len(sa_train), len(sa_test))
+        922 102
+        """
+        from .sframe import SFrame
```
TobyRoseman: I don't think this is going to work either. There is no `.sframe` in GLC. I think you might need to do something like:
https://github.com/dato-code/SFrame/blob/97e13ca91eab82f3a27c65e23316f9617ac0aa99/oss_src/unity/python/sframe/util/__init__.py#L44

`from . import SFrame` I think will work, probably try that one first.

znation: I copied this from another part of the same file: https://github.com/dato-code/SFrame/blob/master/oss_src/unity/python/sframe/data_structures/sarray.py#L698

I am pretty sure this works in both packages since `from .sframe` is a filesystem-relative import and there is an `sframe.py` in the same directory as this file.

TobyRoseman: ah, ok. Sounds good.

ylow:
@znation  there appears to be a merge conflict. Can you fix up and rebase?

Issue #312, opened 2016-05-13
Conversion to Spark dataframe duplicates some rows and looses others


nicoinn:
Hi!

I have noticed a strange bug while converting SFrames to Spark Dataframes: in some rare cases, a few lines of the SFrame become duplicated, while the total number of rows remains the same (thus some get lost). 
The problem happens only when requesting a distributed dataframe. 

Code for reproducing the issue: 

----------------

import sframe
sf_test = sframe.SFrame.read_csv('./data/test.csv')
print len(sf_test)   #Should print  75818
print len(sf_test[sf_test["ID"]==75811])  #Should print 1

import findspark
findspark.init()
import pyspark
sc = pyspark.SparkContext()  
from pyspark.sql import SQLContext
from pyspark.sql.types import *
sqlContext = SQLContext(sc)

spark_test = sf_test.to_spark_dataframe(sc,sqlContext,4)  #<=== If 4 is used, duplication happens, if 1, everything is ok.
sf_test_back = sframe.SFrame(spark_test.toPandas())
print len(sf_test_back) #Should print  75818
print len(sf_test_back[sf_test_back["ID"]==75811])  #Should print 1, prints 2!

----------------

The test.csv file is taken from here: https://www.kaggle.com/c/santander-customer-satisfaction/data

esoroush:
Hi Nicolas, Thanks for reporting this issue. It's a bug deep in our partitioning mechanism. I will let you know when we fix is ready.

TobyRoseman:
@esoroush - can this issue be closed?

esoroush:
This issue is resolved

PR #329, opened 2016-05-25
fix for overlapping partitions

esoroush:


oss_src/sframe/spark_unity.cpp
```diff
@@ -612,11 +612,17 @@ int tordd_main(std::string & _output_directory, size_t & numPartitions, size_t &
...
   auto partition_size = sframe_len/numPartitions;
-  if(partId < (sframe_len % numPartitions)) { 
+  auto remainder = sframe_len % numPartitions;
+  size_t row_start = partition_size * partId; 
+  if(partId < remainder) { 
     partition_size +=1;
+    row_start = partition_size * partId; 
+  } else { 
+    row_start = partition_size * partId + remainder;
   }
```
ylow: This whole logic is really annoying. 
What about something simple like:
```
auto partition_size = sframe_len / numPartitions;
row_start = partition_size * partId;
if (partId != numPartitions - 1) {
    row_end = row_start + partition_size;
} else {
    // last partition
    row_end = sframe_len;
}
```
Yes this has a slight degree of imbalance; but no more than numPartitions off. But the logic is much simpler and easier to show that it is correct.

esoroush:
Hoyt, this is the fix for this issue: https://github.com/dato-code/SFrame/issues/312
whenever we merge this fix, then I can start building for linux/mac and collecting cy_spark_unity.so artifacts and generate a new spark_unity.jar.

oss_src/sframe/spark_unity.cpp
hoytak: Or just do: 

```
row_start = ((part_Id+1) * sframe_len) / num_partitions;
row_end = (part_id * sframe_len) / num_partitions;
```

esoroush: Thanks guys for comments. I changed it to the simpler version.

PR #331, opened 2016-05-25
Use old sort routine for small SFrames. will be faster.

ylow:


oss_src/sframe_query_engine/algorithm/ec_sort.cpp
```diff
@@ -47,9 +47,9 @@ std::shared_ptr<sframe> ec_sort(
     num_rows = infer_planner_node_length(sframe_planner_node);
   }
   ASSERT_GE(num_rows, 0);
-  // fast path for 0 rows.
+  // fast path for few number of rows.
   // fast path for no value columns
-  if (num_rows == 0 ||
+  if (num_rows < 1000 ||
```
hoytak: Do you also want to put a limit on the number of columns?

ylow: Suspect not. I think if there are so few rows, it is *probably* fine. But really. I have no clue what this should be. It is kinda a crappy heuristic for now.

PR #333, opened 2016-05-26
Do not fail silently when cache evicting fails

haijieg:


ylow:
If eviction indeed failed, it is probably a bad idea to release memory. The check has to happen earlier.

ylow:
This is also problematic in that if it fails, it will just keep re-attempting the eviction. The right fix will probably be to completely disable evicting in the event of a failure.

haijieg:
+1

PR #334, opened 2016-05-27
Windows >2GB file write issues

ylow:
Fixes an annoying issue on windows where writes greater than 2GB do not complete.

This is probably due to the underlying windows API being 32-bit.

The solution is to cut up the write buffer and write 1GB at a time.


haijieg:
+1 verified on windows

PR #338, opened 2016-06-01
Revert "Added use of set_new_handler"

ylow:
Reverts dato-code/SFrame#311

This should not have worked. 
https://github.com/dato-code/SFrame/pull/311/files#diff-0d9c5131cf9be17dedddfe3a5cef5aacR258
```
ASSERT_FALSE(mutex.try_lock()) 
```
should always fail because this change changed the mutex to a recursive mutex.

We do indeed have tests that trigger the cache eviction behavior so I have no idea how this ever worked. I will investigate this in greater detail later, but for sanity in master, I am reverting this now.

### DIFF
--- a/oss_src/fileio/fixed_size_cache_manager.cpp
+++ b/oss_src/fileio/fixed_size_cache_manager.cpp
@@ -15,18 +15,6 @@
 namespace graphlab {
 
 namespace fileio {
-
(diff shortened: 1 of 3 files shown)


ylow:
FYI @hoytak

PR #346, opened 2016-06-17
Make Egg Updates

### LINKED ISSUE #344: SFrame usage tracking
Hi 

I wonder if were possible to remove user tracking from SFrame

TobyRoseman:
Stop publishing metrics for `sframe` package. Remove `--prod` flag, it wasn't doing anything but metric publishing. This fixes #344 

Standardize python language version number in file name. This will let PyPI pick the right version of python 3 (i.e. 3.4 or 3.5). It will also prevent PyPI from installing `sframe` on a system with a Python version less than 2.7


znation:
Looks good to me!

PR #359, opened 2016-07-17
zeromq->nanomsg

ylow:


ylow:
welll... the failure was because travis couldn't pull from the repo for the python 2 build... but there are no python changes.

